PR #18, opened 2017-12-30
Docs: Fix reference in example code

georgeclaghorn:
`this.inputElement` doesn’t exist. It’s extracted in the next step. Use the local `target` variable instead.


sstephenson:
💝

PR #26, opened 2018-01-02
Show how multi-word controller names map to files

stevenchanin:
I was integrating Stimulus into a new Rails project and I initially hit some problems figuring out how to get multi-word controller names connected to the files with their behavior (e.g. camelCase, dashes, underscores). Thought this doc PR might help other people coming from Ruby who tend to default to underscores.

### DIFF
--- a/handbook/02_hello_stimulus.md
+++ b/handbook/02_hello_stimulus.md
@@ -69,6 +69,15 @@ Next, we need to tell Stimulus how this controller should be connected to our HT
 
 Identifiers serve as the link between elements and controllers. In this case, the identifier `hello` tells Stimulus to create an instance of the controller class in `hello_controller.js`.
 
+### Multi-word controller names
+If you want to use multiple words for the name of your controller (e.g. it's the "Account Settings" controller), Stimulus will try to connect it with the appropriate file
+
+| data-controller | matching files
+| --------------- | --------------
+| `account-settings` | `account-settings_controller.js`, or `account_settings_controller.js`
+| `accountSettings` | `accountSettings.js`
+| `account_settings` | does not match
+
 ## Is This Thing On?
 
 Reload the page in your browser and you'll see that nothing has changed. How do we know whether our controller is working or not?


rmm5t:
I would have to assume that the _conventional/recommended_ approach would be:

| data-controller | matching file
| --------------- | --------------
| `account-settings` | `account_settings_controller.js`

dwightwatson:
Should there be any consideration to the fact that JavaScript generally uses camel casing for filenames? I feel it's a little out of place to force underscores just because it's inside a Ruby project.

rmm5t:
> _...JavaScript generally uses camel casing for filenames..._

@dwightwatson I understand where you're coming from, but says who?

* https://github.com/aws/aws-sdk-js/tree/master/lib
* https://github.com/node-nock/nock/tree/master/lib
* https://github.com/jprichardson/node-fs-extra/tree/master/lib/copy-sync
* https://github.com/coinbase/gdax-node/tree/master/lib

I think the one of the problems with JavaScript is that conventions don't really exist in the same way as some other communities.

dwightwatson:
Very fair point, I suspect i've just been exposed to camel casing mostly. Having a quick look at other projects shows examples of kebab casing (Prettier) and Pascal casing (React). 

Would it then be better to support a couple of reasonable naming schemes out of the box?

javan:
> I would have to assume that the conventional/recommended approach would be

Your assumption is correct. Our convention is dasherized controller identifiers (`data-controller="account-settings"`) and we technically support underscored or dasherized filenames: `account_settings_controller.js` or `account-settings-controller.js`. Here's the implementation if you're interested: https://github.com/stimulusjs/stimulus/blob/5cf6be6abe39d073596ce657113f15e280ab1d81/packages/%40stimulus/webpack-helpers/index.ts#L16-L37

@stevenchanin, regarding your change. I think this should be documented better in https://github.com/stimulusjs/stimulus/blob/master/INSTALLING.md#using-webpack, not in the handbook.

stevenchanin:
@javan - I must have missed the note in the `using-webpack` section. Based on that, I think we could close this PR.

PR #27, opened 2018-01-02
fix grammatical error in ORIGIN.md

BrianDGLS:
Removes unneeded "an".


javan:
Thank you! Removed the "a" instead :)

PR #29, opened 2018-01-02
Fix its -> it's typo

mikepurvis:


javan:
Fixed in https://github.com/stimulusjs/stimulus/pull/30. Thank you!

PR #40, opened 2018-01-05
Allow to store objects

savroff:
I'm currently building an uploader component and I figured that it would be nice to have a way to store objects in data. For example:

I have an array of values and only one way to manipulate them right now is actually create DOM element per each value.

Where basically I want to have something like this:

```
  set storage(object) {
    const array = [...this.storage, object]
    this.data.set('storage', array)
  }

  get storage() {
    if (this.data.has('storage')) {
      return this.data.get('storage')
    } else {
      return []
    }
  }
```

I don't see a problem with using `JSON` because it has [pretty good benchmarks](https://jsperf.com/json-stringify-versions/10) and I hope people will not manipulate with 200.000 objects here.

Thank you!

### DIFF
--- a/packages/@stimulus/core/src/data_map.ts
+++ b/packages/@stimulus/core/src/data_map.ts
@@ -17,11 +17,13 @@ export class DataMap {
 
   get(key: string): string | null {
     key = this.getFormattedKey(key)
-    return this.element.getAttribute(key)
+    const value = this.element.getAttribute(key)
+    return value ? JSON.parse(value) : null
   }
 
   set(key: string, value): string | null {
     key = this.getFormattedKey(key)
+    value = value ? JSON.stringify(value) : null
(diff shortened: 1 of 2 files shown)


sstephenson:
Thank you so much for the patch!

We’d like to keep Stimulus’ `DataSet` interface string-only to match the DOM [HTMLElement.dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) API. I encourage you to explore your own JSON wrappers on top of that, though.

savroff:
Good, 
thank you!

savroff:
> We’d like to keep Stimulus’ DataSet interface string-only to match the DOM HTMLElement.dataset API. I encourage you to explore your own JSON wrappers on top of that, though.

Also then maybe we can put it somewhere in **Documentation**.

PR #46, opened 2018-01-07
Application#{register,load,unload}

sstephenson:
This branch:

* Introduces the `Definition` interface, which bundles together an identifier and a controller constructor.
* Introduces `Application#load`, which loads one or more `Definition`s, and `Application#unload`, which unloads the controller constructors associated with one or more identifiers. These methods are backed by `Router#loadDefinition` (which does not throw when the identifier is already registered, as `Router#register` did) and `Router#unloadIdentifier`.
* Removes `Application#unregister` and reimplements `Application#register` atop `Application#load`.
* Renames `ContextSet` to `Module`.
* Changes the webpack-helpers API to export a single function, `definitionsFromContext`, which returns an array of `Definition`s which can be passed to `Application#load`.


dixpac:
Love these changes....much nicer API :)

PR #47, opened 2018-01-08
Remove else clause in example

dwightwatson:
Not really sure what your convention on this sort of thing is, but this just simplifies the example by replacing an `else` clause with a regular return.


dhh:
I find spelling out the control flow makes for much nicer reading.

dhh:
(Thanks for taking a moment to propose this, though 🙏)

PR #53, opened 2018-01-10
Add Application.prototype.handleError

javan:
Supply your own error handler function to override the default `console.error`. Useful for sending errors to a tracking service like Sentry.

```js

application.handleError = (error, message, detail) => {
  console.warn(message, detail)
  Raven.captureException(error)
}
```

### DIFF
--- a/packages/@stimulus/core/src/action.ts
+++ b/packages/@stimulus/core/src/action.ts
@@ -48,7 +48,7 @@ export class Action implements EventListenerObject {
     try {
       this.method.call(this.controller, event, event.currentTarget)
     } catch (error) {
-      this.context.reportError(error, `invoking action "${this.descriptor}"`, event, this)
+      this.context.handleError(error, `invoking action "${this.descriptor}"`, { event })
     }
   }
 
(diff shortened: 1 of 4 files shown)


emq:
Welp, that caught me by surprise (obviously in production 😅 )

I would never-ever except such library to swallow errors for me - maybe I ask why `throw` is not the default here? (so if you have sentry it will be reported properly via `window.onerror` _by default_)

And ofc thanks for this PR @javan ! ❤️

sstephenson:
@emq Could you open a new issue so your comment doesn’t get lost in the stream? Thank you!

formigarafa:
It seems the instructions provided in the docs to customize the `handleError` doesn't work. 
The controllers still references the original `handleError` not the substituted/assigned one.

To solve I had to do it like this:
```

const handleErrorOriginal = Application.prototype.handleError
Application.prototype.handleError = function(error, message, detail) {
  console.log("now custom error handler is called");
  handleErrorOriginal.call(this, error, message, detail);
}

```

PR #58, opened 2018-01-13
@stimulus/test

sstephenson:
* Extracts the `@stimulus/test` package which provides a `TestCase` class hierarchy atop QUnit. 
* Adds an `Application.prototype.controllers` getter which returns an array of connected controllers.
* Refactors the existing `@stimulus/core` tests to use `@stimulus/test`.

Example test case:

```typescript
export default class TargetTests extends LogControllerTestCase {
  fixtureHTML = `
    <div data-controller="${this.identifier}">
      <div data-target="${this.identifier}.alpha" id="alpha1"></div>
      <div data-target="${this.identifier}.alpha" id="alpha2"></div>
      <div data-target="${this.identifier}.beta" id="beta1"></div>
    </div>
  `

  "test TargetSet#find"() {
    this.assert.equal(this.findElement("#alpha1"), this.controller.targets.find("alpha"))
  }

  "test TargetSet#findAll"() {
    this.assert.deepEqual(
      this.findElements("#alpha1", "#alpha2"),
      this.controller.targets.findAll("alpha")
    )
  }

  "test TargetSet#findAll with multiple arguments"() {
    this.assert.deepEqual(
      this.findElements("#alpha1", "#alpha2", "#beta1"),
      this.controller.targets.findAll("alpha", "beta")
    )
  }

  // ...
}
```


sstephenson:
CI is currently failing because we need to add a Promise polyfill for IE 11.

javan:
CI's all green now: bbf37abe6f771f2ac550e7f6306376185b581bb3

packages/@stimulus/test/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "main": "dist/module/index.js",
+  "types": "dist/types/index.d.ts",
+  "license": "MIT",
+  "files": [
+    "dist"
+  ],
+  "publishConfig": {
+    "access": "public"
+  }
+}
```
javan: This package should probably depend on `qunit` and `@stimulus/core`

sstephenson: I wasn’t sure how to specify the QUnit dependency, since we don’t import/require it anywhere in this package. 

Adding `@stimulus/core` as a dependency would be cyclical - is that right?

PR #61, opened 2018-01-17
Target definitions

javan:
Define a controller's target names and Stimulus automatically creates properties for accessing them:

```js

  static targets = [ "source", "slide" ]

  initialize() {
    this.sourceTarget  // this.targets.find("source")
    this.sourceTargets // this.targets.findAll("source")

    this.slideTarget  // this.targets.find("slide")
    this.slideTargets // this.targets.findAll("slide")
  }
}
```

### DIFF
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -46,6 +46,27 @@ application.register("hello", HelloController)
 application.register("clipboard", ClipboardController)
 ```
 
+## Using Babel
+
+If you're using [Babel](https://babeljs.io/) with your build system, you'll need to install the [transform-class-properties plugin](https://babeljs.io/docs/plugins/transform-class-properties/) and add it to your configuration:
+
+```js
+// .babelrc
+{
+  "presets": ["env"],
+  "plugins": ["transform-class-properties"]
+}
+```
+
+Or, use the [stage-2 preset](https://babeljs.io/docs/plugins/preset-stage-2/), which includes the transform-class-properties plugin and more:
+
+```js
+// .babelrc
+{
+  "presets": ["env", "stage-2"]
+}
+```
+
 ## Using Without a Build System
 
 If you prefer not to use a build system, you can load Stimulus in a `<script>` tag and it will be globally available through the `window.Stimulus` object.
--- a/handbook/02_hello_stimulus.md
+++ b/handbook/02_hello_stimulus.md
@@ -147,17 +147,19 @@ Stimulus lets us mark important elements as _targets_ so we can easily reference
 > * `hello` is the controller identifier
 > * `name` is the target name
 
-If we pass the target name to the `this.targets.find()` method, Stimulus will return the first matching target element it finds. We can then read its `value` and use it to build our greeting string.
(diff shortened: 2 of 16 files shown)


sstephenson:
https://github.com/stimulusjs/stimulus/blob/ca4f4dcd2de7d9905c7864d52d2497670e15e69a/handbook/02_hello_stimulus.md#L150-L165

Let’s take this change all the way. No need to start by explaining `this.targets`—we can reserve that API for special cases. We should revise the whole section to introduce targets as properties right off the bat.

javan:
Good call. Made some revisions in af233fe50df6b7c799c28c372319ec7ef4554213

sstephenson:
Much better! Probably need to revisit the title and intro, too:

https://github.com/stimulusjs/stimulus/blob/af233fe50df6b7c799c28c372319ec7ef4554213/handbook/02_hello_stimulus.md#L129-L135

“Locate” now seems to describe the implementation rather than the concept. And we aren’t really “referencing” elements “by name” anymore, either. Stimulus is linking targets to controller properties.

sstephenson:
I’m happy with chapter 2 of the handbook now. The last thing we should do is update the installation guide to mention the necessary Babel static properties plugin.

javan:
It turns out that the Babel plugin for class properties is either going to work great or not work at all with our implementation.

## 👍
Add the [transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties/) plugin to your `.babelrc` file:
```json
{
  "plugins": ["transform-class-properties"]
}
```
And classes with static properties like:
```js
class A {
  static targets = […]
}
```
Are compiled to:
```js
var A = function() {}
A.targets = […]
```

## 👎
Use the plugin's `spec: true` option like [webpacker does](https://github.com/rails/webpacker/blob/84298742c9789c986fa64df3812b624bc51376ca/lib/install/config/.babelrc#L16):
```json
{
  "plugins": ["transform-class-properties", { "spec": true }]
}
```
And the same class compiles to:
```js
var A = function() {}
Object.defineProperty(A, "targets", { value: […] })
```
Thus bypassing our setter:
https://github.com/stimulusjs/stimulus/blob/1fcfa4f1567b9d918d9163c662dc67f364f3b296/packages/%40stimulus/core/src/controller.ts#L11-L14

In Babel 7, [this is going to be the default behavior](https://babeljs.io/blog/2017/09/12/planning-for-7.0#stage-3-class-properties-from-stage-2) unless you use the `loose: true` option.

dixpac:
Thinking out loud...is it possible get target properties without specifying them explicitly inside the controller?

```javascript
// src/controllers/hello_controller.js

  //static targets = [ "name" ]       without this line

  greet() {
    const element = this.nameTarget
    const name = element.value
    console.log(`Hello, ${name}!`)
  }
}
```

javan:
@dixpac we could with [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), but it's unsupported in IE and can't be polyfilled.

javan:
Possible solution to the Babel situation: c8e7696746ba706b7dceb76e3c3a355431519733

PR #65, opened 2018-01-20
Namespaced identifiers

sstephenson:
This branch adds support for directory-namespaced controllers in `@stimulus/webpack-helpers`. It does this by mapping `/` in controller filenames to `--`, which makes the identifier safe for use in HTML attributes (and thus compatible with the Stimulus data API).

```js
// controllers/users/list_item_controller.js
  get indexValue() {
    return this.data.get("index")
  }
}
```

```html
<div data-controller="users--list-item" data-users--list-item-index="0">...</div>
```

### DIFF
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -10,7 +10,7 @@ To install Stimulus, add the [`stimulus` npm package](https://www.npmjs.com/pack
 
 Stimulus integrates with the [webpack](https://webpack.js.org/) asset packager to automatically load controller files from a folder in your app.
 
-Use webpack's [`require.context`](https://webpack.js.org/api/module-methods/#require-context) helper in conjunction with Stimulus' `definitionsFromContext` helper to set everything up:
+Call webpack's [`require.context`](https://webpack.js.org/api/module-methods/#require-context) helper with the path to the folder containing your Stimulus controllers. Then, pass the resulting context to the `Application#load` method using the `definitionsFromContext` helper:
 
 ```js
 // src/application.js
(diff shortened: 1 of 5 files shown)


javan:
I like this change. It introduces some subtle differences/requirements that we should at least document:
1. Your `require.context` *must* point to your root controllers directory. Currently you can do `require.context(".", true, /\.js$/)` and it will work the same as `require.context("./controllers", true, /\.js$/)` because we only care about the last segment of the path.
2. You can't use directories just to organize/group your controllers. Essentially the same as the first point. Nested directories have meaning now.

sstephenson:
[Updated the Installation Guide](https://github.com/stimulusjs/stimulus/blob/f7f999a7a4476ece2aebff932b8cf95ae979401e/INSTALLING.md#controller-filenames-map-to-identifiers) to plainly spell out how the mapping works.

etrokal:
I have a doubt about how to use actions and targets identifiers with namespaced controllers. I searched the web a lot, but could not find an answer to this.

So, if I have a controller like this:

```
// controllers/user/sign_in_controller.js
  save() {
    // some functionality
  }
}
```
And in HTML I have this:
```
<div data-controller="user--sign-in">
  // some code
  <button data-action="click->signIn#save">
    Save
  </button>
</div>
```

Is this correct? Don't I need to namespace the action identifier? What about target identifiers? Do they work the same?

skyksandr:
@etrokal, consider the following example:

```js
// controllers/flight_profiles/track_controller.js

  static targets = ['profile', 'place']

  toggle(event) {
    ...
  }
}
```

Template (sorry for HAML):
```haml
%li.flight-profiles__track{data: { controller: 'flight-profiles--track', 'track-id': track.id } }
  .checkbox
    %label
      %input{type: 'checkbox', 'data-action': 'change->flight-profiles--track#toggle'}
```

PR #67, opened 2018-01-22
Introduce a `Controller.bless` hook

sstephenson:
Controllers can override `static bless()` to define additional properties on the module's anonymous subclass.

### DIFF
--- a/packages/@stimulus/core/src/controller.ts
+++ b/packages/@stimulus/core/src/controller.ts
@@ -3,15 +3,21 @@ import { Context } from "./context"
 import { DataMap } from "./data_map"
 import { Scope } from "./scope"
 import { TargetSet } from "./target_set"
+import { defineTargetProperties } from "./target_properties"
 
 export interface ControllerConstructor {
-  targets: string[]
+  import()
   new(context: Context): Controller
 }
 
 export class Controller {
   static targets: string[] = []
+
   readonly context: Context
+
+  static import() {
+    defineTargetProperties(this)
+  }
 
   constructor(context: Context) {
     this.context = context
--- a/packages/@stimulus/core/src/controller.ts
+++ b/packages/@stimulus/core/src/controller.ts
@@ -6,7 +6,7 @@ import { TargetSet } from "./target_set"
 import { defineTargetProperties } from "./target_properties"
 
 export interface ControllerConstructor {
-  import()
+  bless()
   new(context: Context): Controller
 }
 
@@ -15,7 +15,7 @@ export class Controller {
 
   readonly context: Context
 
-  static import() {
+  static bless() {
     defineTargetProperties(this)
   }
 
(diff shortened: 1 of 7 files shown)


packages/@stimulus/core/src/controller.ts
```diff
@@ -3,16 +3,22 @@ import { Context } from "./context"
...
+  import()
   new(context: Context): Controller
 }
 
 export class Controller {
   static targets: string[] = []
+
   readonly context: Context
 
+  static import() {
```
javan: Looking at this externally, I'd expect `import()` to take argument (and import it). How about `setup()` or `prepare()`?

packages/@stimulus/core/src/target_properties.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function getOwnTargetNamesForConstructor(constructor: Function) {
+  const definition = constructor["targets"]
+  return Array.isArray(definition) ? definition : []
+}
+
+function defineLinkedProperties(object: any, properties: PropertyDescriptorMap) {
+  Object.keys(properties).forEach((name) => {
+    if (object[name] === undefined) {
```
javan: This will call a `get`er method if defined, which we probably don't want to do.

screen shot 2018-01-22 at 5 43 25 pm

It's the reason I went with `name in object` in the current implementation: https://github.com/stimulusjs/stimulus/blob/8c141386085d61e4c42593b6b72aa5f8ff7723ed/packages/%40stimulus/core/src/definition.ts#L41

packages/@stimulus/core/src/controller.ts
sstephenson: I picked `import` to match the existing language around `importDefinition`, and for parity with the present-tense lifecycle method names (`initialize`, `connect`, `disconnect`). Fair points though. Of the two I think I prefer `setup`.

packages/@stimulus/core/src/target_properties.ts
sstephenson: 👍 99981ec

packages/@stimulus/core/src/controller.ts
sstephenson: What do you think of the name `bless`?

In Perl, you [`bless` a hash reference to associate it with a class](https://www.perl.com/pub/1999/09/refererents.html/#rule-3-to-create-an-object-bless-a-referent).

This [Stack Overflow answer](https://stackoverflow.com/a/392194) explains it succinctly:

> The effect bless-ing has is that it allows you to apply special syntax to the blessed reference.

It seems fair to say that what we are doing here is applying special properties to a “blessed” controller class.

javan: Love that. #blessed

sstephenson: 🙏

PR #68, opened 2018-01-23
Singular target properties throw an error when the element is missing

sstephenson:
Tired: `undefined is not an object`
Wired: `Missing target element: "clipboard.source"`

This branch fails fast, and with a more helpful error message, when using a `this.fooTarget` linked property. It also adds a new `this.hasFooTarget` boolean linked property for testing whether a target exists.

### DIFF
--- a/packages/@stimulus/core/src/target_properties.ts
+++ b/packages/@stimulus/core/src/target_properties.ts
@@ -16,6 +16,11 @@ export function defineTargetProperties(constructor: Function) {
       get() {
         return this.targets.findAll(name)
       }
+    },
+    [`has${capitalize(name)}Target`]: {
+      get() {
+        return this.targets.has(name)
+      }
     }
   }))
 }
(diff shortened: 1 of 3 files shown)


packages/@stimulus/core/src/target_properties.ts
```diff
@@ -45,3 +55,7 @@ function defineLinkedProperties(object: any, properties: PropertyDescriptorMap)
     }
   })
 }
+
+function capitalize(name: string) {
+  return name.charAt(0).toLocaleUpperCase() + name.slice(1)
```
javan: Any risk of chopping a multibyte char here?

sstephenson: Maybe…? It’d have to be a character outside the BMP which also has a corresponding capital. Seems unlikely enough that this will suffice for now.

PR #110, opened 2018-02-21
TokenObserver

sstephenson:
This branch is a refactoring intended to support future Stimulus features.

* Introduces `TokenObserver` as part of the `@stimulus/mutation-observers` package. A `TokenObserver` watches for token strings in an attribute, asks the delegate to map those tokens to parsed objects, and then notifies the delegate as the objects come and go based on the presence of their underlying tokens.
* Introduces `ScopeObserver` to `@stimulus/core`, which uses `TokenObserver` to instantiate `Scope` objects as identifiers come and go in the `data-controller` attribute.
* Refactors `Router` to use a `ScopeObserver` instead of a `TokenListObserver`.
* Replaces `InlineActionObserver` with `ActionObserver`, which uses `TokenObserver` to instantiate `Action` objects as action descriptors come and go in the `data-action` attribute.
* Refactors `Action` into a value class and moves event logic out into a separate `EventListener` class. Renames `ActionSet` to `EventListenerSet`.
* Improves the error message for action descriptors with missing event names.
* Re-implements the `@stimulus/mutation-observers` test suite with `@stimulus/test`.


packages/@stimulus/polyfills/index.js
```diff
@@ -2,5 +2,6 @@ import "core-js/fn/object/assign"
 import "core-js/fn/set"
 import "core-js/fn/map"
 import "core-js/fn/promise"
+import "core-js/fn/array/find"
```
javan: We’ve been using Array#find all along. Wonder why the tests pass on master currently? https://github.com/stimulusjs/stimulus/blob/01719aff28e775a09699477897bdb0ba64621664/packages/%40stimulus/core/src/inline_action_observer.ts#L81

Either way, we should add it to the list on https://github.com/stimulusjs/stimulus/blob/master/handbook/06_installing_stimulus.md#browser-support

sstephenson: 2ad2cec

PR #120, opened 2018-03-08
[ci skip] Update README for webpacker-helpers

yhirano55:


sstephenson:
Thank you! Could we preserve the anchor?

yhirano55:
@sstephenson Thanks for the comment. Fixed it.

PR #131, opened 2018-03-22
Start when the DOM is interactive

### LINKED ISSUE #97: Embedded script tags prevent target detection
First off, thanks for the great work so far, stimulus is a joy to use.

Not sure if this is a bug or simply a limitation of the browser (or my misunderstanding of mutation observers, or something).

As of stimulus 1.0.0, in my experience if you have a <script> tag within a DOM node marked with a data-controller attribute, the corresponding stimulus controller will find targets with data-target attributes up to the point in the DOM where the <script> tag appears, and then no further. This has actually happened to me twice as I have been using stimulus in places that already had 'sprinkles of jquery', so this may be something that users run into as they replace those parts of their app with stimulus. This likely also comes up if people are using things like cocoon which embed script tags near the form.

Understandably having tiny <script> tags lying around may not be best practice, and I have been able to move them to the end of the document or at least beyond the last data-target, or replace them with stimulus entirely.

Let me know if you cannot reproduce or want help, but I assume this may be a known limitation?

javan:
Alternate title: Defer connecting controllers until the DOM is ready.

Fixes #97


david-a-wheeler:
Does this PR work with turbolinks?  Turbolinks rewrites pages, so its idea of "ready" is different than many other systems.

PR #138, opened 2018-04-02
Upgrade to TypeScript 2.8.1

sstephenson:


package.json
```diff
@@ -26,7 +26,7 @@
     "packages/*",
     "packages/@stimulus/*"
   ],
-  "devDependencies": {
+  "dependencies": {
```
abepark01: Should this still be ‘devDependencies’?

PR #155, opened 2018-05-02
Build system improvements

sstephenson:
* Moves shared config files into a top-level `config/` directory.
* Separates out node and web baseline tsconfig files, and adds a tsconfig extending one or the other to each package.
* Each package is now responsible for cleaning, linting, and building itself. Lerna delegates to package scripts in parallel.
* Resolves a circular dependency with the `@stimulus/test` package by moving `ControllerTestCase` and `ApplicationTestCase` to `@stimulus/core/test` (not publicly exported at this time).
* Adds a top-level `yarn watch` command to build in response to changes for all packages in the repository.


javan:
Looks great! One last thing: We should exclude all of the `dist/test/` dirs so they're not published to npm.

PR #161, opened 2018-06-01
Mutation observers: Avoid adding disconnected elements

javan:
This change works around a somewhat unexpected `MutationObserver` behavior: When a node is removed from an observed tree *and then* another node is added to it synchronously, the observer will include that added node in its next batch of mutation records. Example: https://jsfiddle.net/javan/acxd6v1x/

So, if you remove an element from the DOM and then add a `data-controller` element to it, Stimulus will match the added element and install a new controller. Because the element is already disconnected from the DOM, Stimulus will never *unmatch* and `disconnect()` its controller. AKA a memory leak. Example: https://jsfiddle.net/javan/vqht4n5s/

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "tslint": "^5.7.0",
     "typedoc": "^0.11.1",
     "typedoc-plugin-stimulus": "file:packages/typedoc-plugin-stimulus",
-    "typescript": "^2.8.1",
+    "typescript": "^2.8.4",
     "webpack": "^3.10.0"
   }
 }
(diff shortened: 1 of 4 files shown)


sstephenson:
Do we need an `isConnected` polyfill?

javan:
The `isConnected` test is an optimization.
https://github.com/stimulusjs/stimulus/blob/d3ea33716e2315ae3a1c0189e61bb1e136eccd82/packages/%40stimulus/mutation-observers/src/element_observer.ts#L132-L136
Browsers that do support it will return a boolean, and if there's a mismatch we can skip the slightly more expensive `element.contains()` call in the `else` branch. Browsers that don't support will return `undefined` and jump straight to the `else` branch.

javan:
A  polyfill might look something like this, which would end up being even more work:
```js
if (!("isConnected" in Node.prototype)) {
  Object.defineProperty(Node.prototype, "isConnected", {
    get() {
      return document.documentElement.contains(this)
    }
  })
}
```
(not that it really matters)

PR #178, opened 2018-08-18
Add clarification on how events are supported

ProtonScott:
So I have been working my way through adding stimulus to my application, and I stumbled upon the fact that it was actually somewhat unclear (to me, perhaps not others) how to support triggering other events.

Now I understand on the back side from what I can tell you're implementing the addEventListener API so that makes this easy, so I did figure out that you're simply short forming event->controller#method to the equivalent addEvenListener API. 

However, it did take me digging into the code a little to surmise this properly. I want to add a blurb to the docs that will make this clear and apparent.

The example is an example I found from an earlier issue, [issue 127](https://github.com/stimulusjs/stimulus/issues/127)

If its too verbose or whatever, let me know, but I think it would be great if the docs explicitly stated that this is how it works. Explicit vs implicit and all that jazz.

### DIFF
--- a/handbook/03_building_something_real.md
+++ b/handbook/03_building_something_real.md
@@ -131,6 +131,14 @@ export default class extends Controller {
 }
 ```
 >
+> Window events are supported to, for instance, this commonly deployed method for linking YouTube videos onto a site:
+
+```html
+<div data-controller="youtube" data-action="youtube-api:ready@window->youtube#init_player">
+  <div id="player">
+</div>
+```
+>
 
 Finally, in our `copy()` method, we can select the input field's contents and call the clipboard API:
 
--- a/handbook/03_building_something_real.md
+++ b/handbook/03_building_something_real.md
@@ -101,6 +101,36 @@ We want a click on the button to invoke the `copy()` method in our controller, s
 > input type=submit | click
 > select            | change
 > textarea          | change
+>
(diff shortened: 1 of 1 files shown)


danielcompton:
I like this, I had the same 'aha' moment too as I was learning Stimulus.

skyksandr:
Also it worth to mention you can listen to events on window with following form:
```
<div data-controller="youtube" data-action="youtube-api:ready@window->youtube#init_player">
  <div id="player">
</div>
```

handbook/03_building_something_real.md
```diff
@@ -101,6 +101,44 @@ We want a click on the button to invoke the `copy()` method in our controller, s
...
+    event.preventDefault()
+  }
+
+  dropCard() {
+    // ...
+  }
+}
+```
+>
+> Window events are supported to, for instance, this commonly deployed method for linking YouTube videos onto a site:
```
adamwiggall: Typo: Window events are supported to -> ... events are supported **too**.

sstephenson:
Thanks very much for the pull request! We've been working on new documentation for the 1.1 release to cover topics like these.

I believe everything here is covered in our [reference chapter on actions](https://stimulusjs.org/reference/actions), but please do open new issues if there's something we missed.

PR #182, opened 2018-08-26
Use dot instead of double dash for clear namespacing

gauravtiwari:
If we have controllers like so: 

```yml
controllers: 
  admin: 
      - reference_controller.js 
  app: 
      - reference_controller.js 
```

```js
const context = require.context('../controllers', true, /\.js$/)
```

It's not clear how to add them to views.

After some digging, I figured that we could do this `admin--reference` which is unintuitive and inconsistent since we are using `.` notation for `target` and `action` referencing. 

```html
<div data-controller="admin--reference.someTarget"></div>
<div data-controller="app--reference"></div>
```

This PR changes `--` to `.` to support namespaced controllers. 

--- 

Also, I am wondering, 

1. Do we need a `_controller` suffix? Wouldn't it be easier if it's just called `something.js` and then perhaps we can verify after module load that it's indeed a stimulus controller instance? 

2. If we have a layout like this: 

```
reference: 
      - reference.sass
      - _reference.html.erb
      - reference_controller.js 
```

Then, the identifier becomes repetitive: 

<div data-controller="referencee.reference.target"></div>
<div data-controller="reference.reference"></div>

So, would it makes sense we make controllers globally unique and use only the filename as an identifier?

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -44,7 +44,7 @@ For example, this element has a controller which is an instance of the class def
 <div data-controller="reference"></div>
 ```
 
-If you have controllers within a namespace for example: 
+If you have controllers under a namespace: 
 
 ```yml
 controllers: 
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -44,6 +44,28 @@ For example, this element has a controller which is an instance of the class def
 <div data-controller="reference"></div>
 ```
 
+If you have controllers within a namespace for example: 
+
+```yml
+controllers: 
+	admin: 
+		- reference_controller.js 
+	app: 
+		- reference_controller.js 
+```
+
+```js
+const context = require.context('../controllers', true, /\.js$/)
+application.load(definitionsFromContext(context))
+```
+
+Then, you can reference them in your views like so: 
+
+```html
+<div data-controller="admin.reference"></div>
+<div data-controller="app.reference"></div>
+```
+
 ## Scopes
 
 When Stimulus connects a controller to an element, that element and all of its children make up the controller's _scope_.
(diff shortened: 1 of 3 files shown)


javan:
> After some digging, I figured that we could do this `admin--reference` which is unintuitive and inconsistent since we are using `.` notation for `target` and `action` referencing.

That's not true. We use `.` for targets, `#` for actions, and `--` for controller namespacing. So we're consistent in that each uses a different, visually distinct syntax. Using `.` for two out of three makes the HTML less readable, IMO.

> Do we need a `_controller` suffix? Wouldn't it be easier if it's just called `something.js` and then perhaps we can verify after module load that it's indeed a stimulus controller instance?

Would you say the same about `app/controllers/*` files in Rails?

gauravtiwari:
@javan 

> That's not true. We use . for targets, # for actions, and -- for controller namespacing. So we're consistent in that each uses a different, visually distinct syntax. Using . for two out of three makes the HTML less readable, IMO.

I see, forgive me, didn't see this part documented anywhere, that's why felt this might not be officially supported/considered. 

In regards to readability, I feel the `.` dot is more readable and cleaner but perhaps that's just a preference. For example: `admin--analytics--reference` vs `admin.analytics.reference`, the latter aligns well with JS and ruby dot notation too.

  > Would you say the same about app/controllers/* files in Rails?

😄 No, I wouldn't that's Rails specific. Sorry, I made the PR in rush and didn't properly explained the reasoning. Basically,  I felt that a stimulus controller is POJO, which is great and that's why felt that it might be better if the library didn't make any assumption on how a file is named as long as its `.js`. 

**For example:** React class inheritance style looks very similar to Stimulus but it doesn't enforce using `_component` or `.component` as a suffix. 

```js
class Food extends React.Component {}
// foo.js

class Bar extends Stimulus.Controller {}
// foo_controller.js
```

Anyway, I was just curious  😄 Feel free to close

sstephenson:
The `--` separator isn't arbitrary—identifiers have to be representable as HTML data attributes, and underscores and dots are unidiomatic in attribute names.

Either way, nothing _prevents_ you from using a different namespace separator. You can pass whatever identifier you want to `Application#register()`. This is just the convention we use in our webpack autoload helpers.

gauravtiwari:
Thanks, @sstephenson 

Sure, makes sense 👍

PR #125, opened 2018-03-13
Add "security" chapter to handbook

david-a-wheeler:
Add a chapter to explain how to write more secure software
using Stimulus.  This information isn't obvious, it's important,
and it's pretty straightforward.

This information was provided in part by Sam Stephenson
on Tue, 13 Mar 2018 16:35:18 +0000. See
https://discourse.stimulusjs.org/t/security-csp-and-stimulus/171/2

Signed-off-by: David A. Wheeler <dwheeler@dwheeler.com>

PR #184, opened 2018-08-29
Add a "Security Considerations" document to the repo root

sstephenson:
This branch pulls in answers to the questions posed by @david-a-wheeler in this thread: https://discourse.stimulusjs.org/t/security-csp-and-stimulus/171

I think it's good to publish these answers, but after some consideration I don't think they need to be elevated to the same level as controllers, actions, and targets in our [current](https://stimulusjs.org/handbook/introduction) [documentation](https://stimulusjs.org/reference/controllers) site (#125).

Instead, I've compiled them into a SECURITY.md file at the root of the repository, where they should be obvious if you're looking for them, and less alarming if not.

/cc @jeremy

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,10 +1,10 @@
 # Security Considerations
 
 ### Q: Can I be confident that if my cross-site scripting countermeasures fail, there is no way for an attacker to run arbitrary JavaScript using Stimulus?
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Having information somewhere is definitely better than *not* having it.  I think the tutorial should at least reference this information (SECURITY.md).  The number of developers who should care about security is basically 100%, so this information really shouldn't be hiding.  Exactly *where* it goes is, to me, less important.  I'll file a separate issue regarding "there should be a link".

PR #190, opened 2018-09-03
Remove suggestion to use stage-2 preset from handbook

rmacklin:
Babel 7 has been released, and the stage presets have been deprecated/discontinued:

https://babeljs.io/blog/2018/08/27/7.0.0#major-breaking-changes
https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets

Since Babel no longer recommends/supports using the stage presets, I think it makes sense to remove this from the Stimulus handbook.


javan:
Thanks!

PR #195, opened 2018-09-24
Update controllers.md (documentation)

richjdsmith:
Not sure if it is wanted, but I know examples help me understand and this was something I had to take an extra moment to understand. The PR is for an example on using multi-word controllers.


sstephenson:
Thank you for the pull request! I'm not sure it deserves the extra prominence of having its own section and example, but we'll keep an eye out for confusion and take a look at other ways we can demonstrate these conventions.

danielcompton:
FWIW some sort of reference section on naming (either here or elsewhere) would be useful for me. I've found it tricky sometimes to know how to line up my naming conventions.

danielcompton:
Awesome, sorry for missing this.

PR #202, opened 2018-11-02
The Values and Classes APIs

sstephenson:
This pull request introduces two new APIs to Stimulus: _Values_ and _Classes_. These APIs are designed to improve upon, and ultimately obviate, the current [Data Map API](https://stimulusjs.org/reference/data-maps). We plan to ship them together in the upcoming Stimulus 2.0 release.

## Values

Most uses of the Data Map API in Basecamp fall under the following categories:

* Storing small strings, such as URLs, dates, or color values
* Keeping track of a numeric index into a collection
* Bootstrapping a controller with a JSON object or array
* Conditioning behavior on a per-controller basis

However, the Data Map API only works with string values. That means we must [manually convert to and from other types](https://stimulusjs.org/handbook/managing-state#reading-initial-state-from-the-dom) as needed. The Values API handles this type conversion work automatically.

### Value properties

The Values API adds support for a static `values` object on controllers. The keys of this object are Data Map keys, and the values declare their data type:

```js
  static values = {
    url: String,
    refreshInterval: Number,
    loadOnConnect: Boolean
  }

  connect() {
    if (this.loadOnConnectValue) {
      this.load()
    }
  }

  async load() {
    const response = await fetch(this.urlValue)
    // ...
    setTimeout(() => this.load(), this.refreshIntervalValue)
  }
}
```

### Supported types and defaults

This pull request implements support for five built-in types:

Type | Serialized attribute value | Default value
---- | -------------------------- | -------------
Array | `JSON.stringify(array)` | `[]`
Boolean | `boolean.toString()` | `false`
Number | `number.toString()` | `0`
Object | `JSON.stringify(object)` | `{}`
String | Itself | `""`

Each type has a default value. If a value is declared in a controller but its associated data attribute is missing, the getter property will return its type's default.

### Controller properties

Stimulus automatically generates three properties for each entry in the object:

Type | Kind | Property name | Effect
---- | ---- | ------------- | -----------
Boolean, Number, Object, String | Getter | `this.[name]Value` | Reads `data-[identifier]-[name]-value`
Array | Getter | `this.[name]Values` | Reads `data-[identifier]-[name]-values`
Boolean, Number, Object, String | Setter | `this.[name]Value=` | Writes `data-[identifier]-[name]-value`
Array | Setter | `this.[name]Values=` | Writes `data-[identifier]-[name]-values`
Boolean, Number, Object, String | Existential | `this.has[Name]Value` | Tests for presence of `data-[identifier]-[name]-value`
Array | Existential | `this.has[Name]Values` | Tests for presence of `data-[identifier]-[name]-values`

Note that array values are always pluralized, both as properties and as attributes.

### Value changed callbacks

In addition to value properties, the Values API introduces _value changed callbacks_. A value changed callback is a specially named method called by Stimulus whenever a value's data attribute is modified.

To observe changes to a value, define a method named `[name]ValueChanged()`. For example, a slideshow controller with a numeric `index` property might define an `indexValueChanged()` method to display the specified slide:

```js
  static values = { index: Number }

  indexValueChanged() {
    this.showSlide(this.indexValue)
  }

  // ...
}
```

Stimulus invokes each value changed callback once when the controller is initialized, and again any time the value's data attribute changes.

Even if a value's data attribute is missing when the controller is initialized, Stimulus will still invoke its value changed callback. Use the existential property to determine whether the data attribute is present.

<br>

## Classes

Another common use of the Data Map API is to store CSS class names.

For example, Basecamp's copy-to-clipboard controller applies a CSS class to its element after a successful copy. To avoid inlining a long [BEM string](https://css-tricks.com/bem-101/) in our controller, and to keep things loosely coupled, we declare the class in a `data-clipboard-success-class` attribute:

```html
<div data-controller="clipboard"
     data-clipboard-success-class="copy-to-clipboard--success">
```

and access it using `this.data.get("successClass")` in the controller:

```js
this.element.classList.add(this.data.get("successClass"))
```

The Classes API formalizes and refines this pattern.

### Class properties

The Classes API adds a static `classes` array on controllers. As with [targets](https://stimulusjs.org/reference/targets#definitions), Stimulus automatically adds properties for each class listed in the array:

```js
// clipboard_controller.js

  initialize() {
    if (/* ... */) {
      this.element.classList.add(this.supportedClass)
    }
  }

  copy() {
    // ...
    this.element.classList.add(this.successClass)
  }
}
```

Kind | Property name | Effect
---- | ------------- | -----------
Getter | `this.[name]Class` | Reads the `data-[identifier]-[name]-class` attribute
Existential | `this.has[Name]Class` | Tests whether the `data-[identifier]-[name]-class` attribute is present

#### Declarations are assumed to be present

When you access a class property in a controller, such as `this.supportedClass`, you assert that the corresponding data attribute is present on the controller element. If the declaration is missing, Stimulus throws a descriptive error:

Screenshot showing error message: "Missing attribute 'data-clipboard-supported-class'"

If a class is optional, you must first use the existential property (e.g. `this.hasSupportedClass`) to determine whether its declaration is present.

<br>

----

## Unifying target attributes

We've made a change to the target attribute syntax to align them with values and classes, and also to make the controller identifier more prominent by moving it into the attribute name.

The original syntax is:

```html
<div data-target="[identifier].[name]">
````

and the updated syntax is:

```html
<div data-[identifier]-target="[name]">
```

### The original syntax is supported but deprecated

Stimulus 2.0 will support both syntaxes, but using the original syntax will display a deprecation message in the developer console. We intend to remove the original syntax in Stimulus 3.0.

----

### Try it out in your application

Update the Stimulus entry in `package.json` to point to the latest development build:

```
"stimulus": "https://github.com/stimulusjs/dev-builds/archive/b8cc8c4/stimulus.tar.gz"
```


PFWhite:
Very excited about the values api! Both these features seem really similar in that they both: get declarations from a static array, read the DOM, then add properties to the controller instance, just like the data-target api. Maybe the more general API could be exposed? It might be nice to have the ability to put custom data-attributes and be able to write handlers.

sstephenson:
> Maybe the more general API could be exposed? It might be nice to have the ability to put custom data-attributes and be able to write handlers.

Could you clarify what you mean, exactly?

adrienpoly:
That is awesome, I love the new value API. Have been doing similar things in some controllers. I am glad this can be now fully refactored with this API. Will try to test it over the WE on one particular project.

One question for the value changed callbacks: Would it be possible to have a global `valueChanged()` callback (or `anyValueChanged()`). 

Let's say that I have 3 value attributes. I would like to have only one callback function that would be triggered whenever any of the 3 value change (ideally with the changed value in the params)?

Will need to test with the current API but my initial thinking was that this could be useful. If I come up with a real use case, I will post it back.

sstephenson:
> Would it be possible to have a global `valueChanged()` callback (or `anyValueChanged()`).
> 
> Let's say that I have 3 value attributes. I would like to have only one callback function that would be triggered whenever any of the 3 value change (ideally with the changed value in the params)?
> 
> Will need to test with the current API but my initial thinking was that this could be useful. If I come up with a real use case, I will post it back.

Yes, that's definitely possible, and seems intuitively appealing too. Probably something we'd want to add in a subsequent release. Please do share your use case if you come up with one!

sstephenson:
> So the values feature covers: json, strings, numbers, and arrays. What if I had something like binary data from some custom protocol. I could then have a "data-blob" property in the HTML, a static "blobs" array in the controller definition, and then some handler which would get passed those attributes on component intialization, which would decode my binary object and add the getters and settes for "this.somethingBlobs".

Yeah, we talked about the possibility of supporting custom types when designing the Values API. It's not too hard to imagine something like:
```js
class extends Controller {
  static values = { person: Person }
}
```
which would automatically serialize and deserialize into an instance of your `Person` class. I think that's something that'd be interesting to explore for a future release.

That said, I don't think you'd ever want to store binary data like that in an HTML attribute. Probably best to stick with a URL instead.

> If this was available people could have written data-number, and declared a numbers static array and had some handler that would split the array and add the object getters and setters on their own.

Hmm, I don't immediately see how exposing a single attribute of each type, and requiring you to pack and unpack objects manually into that attribute, would be an improvement over what's here... maybe you could show a HTML and JavaScript code example that illustrates your use case?

tvandervossen:
My first feeling upon seeing this is that `data-classes` might work better as an attribute name than `data-class`. This makes it match the naming of the static `classes` array on the Stimulus controller and it also avoids newcomers and other peopler reading existing template source from thinking the `data-class` attribute is somehow related to the `class` attribute on the same element.

adrienpoly:
I have tested the value API with the flatpickr wrapper I developed and it works great.

I have encountered a small issue.
one of the options of flatpickr is `time_24hr` (with an `underscore`)

In Rails I have something like that 
```
<%= f.text_field :start_at,
    data: {
      controller: "flatpickr",
      flatpickr_time_24hr: true,
    } %>
```
and it generate in the HTML `data-flatpickr-time-24hr`

but the dasherize function in stimulus does not dasherize `_`. 

so my `static values = { time_24hr: Boolean }` is not recognized. This create the following functions: `hasTime_24hrValue` and `time_24hrValue`

I am really unsure whether it is a Stimulus issue or more a convention issue at the first place when putting an underscore before numbers in a JS variable. 
On my side I ended up modifying my dasherize funtion to convert underscore to dash.

my 2 cents

PFWhite:
@sstephenson 
> automatically serialize and deserialize into an instance of your `Person` class. I think that's something that'd be interesting to explore for a future release.
> 
Great! That's an excellent idea although I think that is still covered by my more general feature request.


> That said, I don't think you'd ever want to store binary data like that in an HTML attribute. Probably best to stick with a URL instead.
> 
You're right, the example was bad.


> Hmm, I don't immediately see how exposing a single attribute of each type, and requiring you to pack and unpack objects manually into that attribute, would be an improvement over what's here... maybe you could show a HTML and JavaScript code example that illustrates your use case?

It isn't better in that case. But clearly you all at basecamp thought that having a specific way to namespace classes was important enough to make this feature and serialization scheme with `data-class`. You could have done it any number of ways, static JSON that's available to your controllers, subclassing the stimulus controller and writing a css-class deserializer in the connect method, etc.... 

But instead you all changed the stimulus core code, and made a special data attribute that gets handled automatically in the lifecycle of all components. I instead would like to:

1. indicate to stimulus that `data-something` should be handled for all components at creation time (or whenever classes, values, targets, actions happens)
2. define a `somethingHandler` that should be passed the instantiated class, and the value of the attribute so I can parse the data and do something to the instance before connect happens.

If this were available, the values and classes api just become flavors of this more general feature, in fact, a whole suite of possible features become just particular implementations. One could have a `data-subscribe` which would help the controller communication issue that pops up here often. Or a `data-person` that calls your Person constructor like you mentioned.

PFWhite:
@sstephenson Read through core, now I realize I need to override initialize method. I missed it on [here](https://stimulusjs.org/reference/lifecycle-callbacks) because it doesn't have a section like connect and disconnect do. This does everything I need it to.

sstephenson:
> My first feeling upon seeing this is that `data-classes` might work better as an attribute name than `data-class`. This makes it match the naming of the static `classes` array on the Stimulus controller and it also avoids newcomers and other peopler reading existing template source from thinking the `data-class` attribute is somehow related to the `class` attribute on the same element.

`data-classes` is one alternative attribute name we've discussed.

All of our other attribute names are singular, including `data-target`, which maps to a plural static array property named `targets`. [Token list](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList) attributes with singular names (`data-controller`, `data-action`, `data-target`, `data-class`) are consistent with HTML (`class`, `rel`, `sandbox`).

However, I agree `data-class` could be misperceived as having the same behavior as the HTML `class` attribute. I wonder if `data-classes` is possibly too subtle of a difference, then?

The other name we've tossed around is `data-class-map` ([ClassMap](https://github.com/stimulusjs/stimulus/blob/4b67d3a044d955fdfad8366b2cc1f18c3e0532f1/packages/%40stimulus/core/src/class_map.ts) is the name of the object that decodes these attributes).

dhh:
I don't see how anyone is going to be confused that data-class does not
mean the same as class. And if so, data-classes is in the same domain, but
breaks parity with the rest of the attributes, as well as the controller
based access methods. Strong vote to stick with data-class.

sstephenson:
> I have tested the value API with the flatpickr wrapper I developed and it works great.

Wonderful! Thanks for testing it out ✌️

> I have encountered a small issue.
> one of the options of flatpickr is `time_24hr` (with an `underscore`)
> 
> In Rails I have something like that
> 
> ```
> <%= f.text_field :start_at,
>     data: {
>       controller: "flatpickr",
>       flatpickr_time_24hr: true,
>     } %>
> ```
> and it generate in the HTML `data-flatpickr-time-24hr`
> 
> but the dasherize function in stimulus does not dasherize `_`.
> 
> so my `static values = { time_24hr: Boolean }` is not recognized. This create the following functions: `hasTime_24hrValue` and `time_24hrValue`
> 
> I am really unsure whether it is a Stimulus issue or more a convention issue at the first place when putting an underscore before numbers in a JS variable.
> On my side I ended up modifying my dasherize funtion to convert underscore to dash.

In general, keys should be snake_case in Ruby, kebab-case in HTML, and camelCase in JavaScript, following each language's own conventions. Rails and Stimulus attempt to convert between these as necessary. That means no underscores in your Stimulus controllers.

You've hit a case where the mapping between the three is not "round-trippable". Your snake-case data key `time_24hr` becomes the kebab-case attribute name `time-24hr`. But in camelCase, you'd write `static values = { time24hr: Boolean }`, which translates to the kebab-case attribute name `time24hr`. Blech!

To fix this, I'd suggest picking a different name that doesn't have this problem. For example, by moving the `24hr` before `time`, you get the round-trippable `data_flatpickr_24hr_time: true` — `data-flatpickr-24hr-time` — `static values = { 24hrTime: Boolean }`.

sstephenson:
Actually, I think we can do better. By replacing [this line](https://github.com/stimulusjs/stimulus/blob/4b67d3a044d955fdfad8366b2cc1f18c3e0532f1/packages/%40stimulus/core/src/class_properties.ts#L15) with:
```js
  const name = `${camelize(key)}Class`
```
you'd be able to specify non-roundtrippable `static values` keys in kebab-case, just as they appear in HTML:
```js
class extends Controller {
  static values = { "time-24hr": Boolean }
}
```
while still being able to use the camelized form for property access: `this.time24hrValue`.

sstephenson:
Thanks for pointing this out, @adrienpoly! fd71f88

tvandervossen:
> All of our other attribute names are singular […]

Sorry for overlooking that. Using `classes` would indeed be inconsistent.

> I don't see how anyone is going to be confused that data-class does not mean the same as class.

I’m not worried about people assuming it means the exact same thing; it just feels a bit generic and too close to the naming of the regular class attribute.

What seems really nice about the existing attribute names (`data-controller`, `data-action`, and `data-target`) is that even without any knowledge about Stimulus, someone reading HTML source can figure out what things are and how they’re going to behave.

> The other name we've tossed around is `data-class-map` (ClassMap is the name of the object that decodes these attributes).

Have to say I’d prefer to see that while reading HTML source, but in the end it might indeed be more sensible to have the attribute naming match the property naming in the controller.

javan:
What I dislike about `data-class` is the word “class”, which is overloaded with meaning. Controllers are classes, CSS has classes, elements have a `class` attribute, but no `class` *property* (they have `element.className`, `element.classList` properties).

Seeing (somewhat contrived) HTML like,
```html
<div data-controller="publication" data-class="published:post">
```
I’m reminded of ActiveRecord’s syntax for declaring an associations class:
```ruby
has_many :published_posts, class_name: "Publication::Post"
```
It’s not immediately clear to me that the `data-class` value… 
1. is a key/value pairing
1. contains CSS class names
1. isn’t specifying a JavaScript `class` for the controller

---

adrienpoly:
> In general, keys should be snake_case in Ruby, kebab-case in HTML, and camelCase in JavaScript, following each language's own conventions. Rails and Stimulus attempt to convert between these as necessary. That means no underscores in your Stimulus controllers.

I know! here as I am doing a wrapper between Flatpickr and Stimulus, I am taking the options that are provided by the parent library. I agree that time_24hr was probably a bad choice at the first place.

> You've hit a case where the mapping between the three is not "round-trippable". Your snake-case data key `time_24hr` becomes the kebab-case attribute name `time-24hr`. But in camelCase, you'd write `static values = { time24hr: Boolean }`, which translates to the kebab-case attribute name `time24hr`. Blech!

Thanks! just learned a new word and concept `round trippable` excellent 

> Thanks for pointing this out

great : Will test the new build pack

javan:
> how about for the class `data-class-key="value"`

Like this?
```html
<div data-controller="clipboard"
     data-class-success="copy-to-clipboard--success"
     data-class-supported="copy-to-clipboard--supported">
```

ngan:
Can't the new Value API do what the Class API is doing?

```
<div data-controller="clipboard" data-clipboard-success-class="copy-to-clipboard--success">
```

Then access it with:
```
this.successClassValue
```

Just my 2 cents, but I'm not sure what the gain is for specifically pulling out css classes as another concept.

alinnert:
I'm using Stimulus 1.1 in a project that we're building right now. I'm in a team with a back-end developer (while I'm responsible for the client side stuff: HTML, JS, Sass) and he told me that the HTML "I generate" is very *talkative* i. e. it's too much code - although he agrees that it's a good thing that you have full control over every aspect of the HTML.

Based on this I do think that combing all classes into one property is a very good idea, because it reduces the source code significantly (and we're also using BEM class names, like in the examples above). I also think that `data-class-success="copy-to-clipboard--success"` doesn't add much value over the current Data Map feature. So I'm totally in for a `data-class(-map)` attribute.

alinnert:
May I throw in an idea about the values API - based on the idea about custom value types? I'd prefer an API like this:

~~~ js
  static values = {
    url: String, // Shortcut for `url: { type: String }`?
    loadItems: { type: Number, default: 5 }
  }
}
~~~

This is similar to Vue's API. That way you can define your own defaults. The above could be an example for a "load more" button below a list of articles or search results of some kind that loads 5 items by default when the user clicks on the button. This can later also be extended to support custom serialize/deserialize functions (which I prefer over passing classes) and other stuff (mark a value as required maybe? ← not necessarily something I need, just an idea).

Example for a Controller that displays a FontAwesome icon:

~~~ html
<div data-controller="icon" data-icon-name="bookmark regular"></div>
~~~

~~~ js
export class IconController extends Controller {
  static values = {
    name: {
      type: String,
      required: true,
      deserialize (stringValue) {
        const [name, style = 'solid'] = stringValue.split(' ')
        return { name, style }
      },
      serialize ({ name, style = '' } = {}) { return (`${name} ${style}`).trim() }
    }
  }

  someMethod () {
    const { name, style } = this.nameValue
    name === 'bookmark'
    style === 'regular'
  }
}
~~~

(I do think that `parse` and `stringify` are very valid alternative names for `serialize` and `deserialize` though. Similar to the methods in `JSON`.)

What do you think about this?

mrhead:
I really like both new APIs 👏.

However, I'm not sure how to set values via JSON. Is something like the following possible? I tried different variations, but it wasn't working for me:

```js
  static values = {
    price: Number,
    discountedPrice: Number,
    // a few other values
  }
}
```

```html
<div data-controller="checkout" data-checkout-values="<%= @checkout.to_json %>">
  ...
</div>
```

If there are a lot of values, then I find it cumbersome to manually create an HTML attribute for each one of them.

I know that I can do the following, but then I'm not taking the advantage of the new values API:

```js
  static values = {
    checkout: Object
  }
}
```

```html
<div data-controller="checkout" data-checkout-checkout="<%= @checkout.to_json %>">
  ...
</div>
```

Thanks!

bruncun:
The new APIs are great! Love the terseness. What do y'all think about adding [hooks](https://reactjs.org/docs/hooks-intro.html)? They could help with reusing/organizing logic, sharing state, and understanding complex components. Implementation-wise, a controller's values could be safely exposed to other controllers by attaching getters/setters to action events. Here's a [proof of concept](https://stimulus-hooks.glitch.me) that mirrors the one from the React docs:

**index.html**
`
<body data-controller="count example" data-count-value="0">
  <p data-target="example.text">You clicked 0 times</p>
  <button data-action="count#use example#increment">Click me</button>
</body>
`

**count_controller.js**
`
 static values = { 
    value: Number
  }

  use(e) {
    e.count = this.valueValue;
    e.setCount = this.setCount.bind(this);
  }

  setCount(newValue) {
    this.valueValue = newValue;
  }

  valueValueChanged() {
    document.title = `You clicked ${this.valueValue} times`;
  }
}
`

**example_controller.js**
`
  static targets = ["text"]

  increment({ count, setCount }) {
    count = count + 1;
    this.textTarget.innerText = `You clicked ${count} times`;
    setCount(count);
  }
}
`

jmas:
> I really like both new APIs 👏.
> 
> However, I'm not sure how to set values via JSON. Is something like the following possible? I tried different variations, but it wasn't working for me:
> 
> ```js
> export default class extends Controller {
>   static values = {
>     price: Number,
>     discountedPrice: Number,
>     // a few other values
>   }
> }
> ```
> 
> ```
> <div data-controller="checkout" data-checkout-values="<%= @checkout.to_json %>">
>   ...
> </div>
> ```
> 
> If there are a lot of values, then I find it cumbersome to manually create an HTML attribute for each one of them.
> 
> I know that I can do the following, but then I'm not taking the advantage of the new values API:
> 
> ```js
> export default class extends Controller {
>   static values = {
>     checkout: Object
>   }
> }
> ```
> 
> ```
> <div data-controller="checkout" data-checkout-checkout="<%= @checkout.to_json %>">
>   ...
> </div>
> ```
> 
> Thanks!

My opinion that this approach will break value changes tracking. MutationObserver API is tracking attribute changes, but when we store all data as object in one attribute - any change in data object will produce all values changes (all `xxxValueChanged()` methods will be invoked).

adammiribyan:
@sstephenson I found myself using the class attributes for specifying the selectors outside the controller:

```html
<div data-controller="highlight-comment"
data-class="container:.panel.clozeable modalTrigger:.panel-footer>button"></div>
	
```

and then doing something like:

```js
  static classes = [ 'container', 'modalTrigger' ]

  connect() {
    if (this.shouldHighlightComment) {
      const commentNode = document.querySelector(`#clozeable_comment_${this.commentIdFromURL}`)
      commentNode.closest(this.containerClass).querySelector(this.modalTriggerClass).click()
    }
  }
}
```

Which lead me to thinking the `data-class` might not be the best term for this. My two cents are to consider `matchers` or `selectors`. Hope this helps!

jmas:
What about `classNames`?

As we already have in JS:
```js
element.className = 'foo bar';
```

Controller example:
```js
class extends Controller {
    static classNames = ['disabled', 'active'];

    set active(isActive) {
        this.element.classList.toggle(this.activeClassName, isActive);
    }

   set disable(isDisabled) {
        this.element.classList.toggle(this.disabledClassName, isDisabled);
   }
}
```

Template example:
```html
<button
    data-controller="button"
    data-class-names="button.is-active:active button.is-disabled:disabled"
>Send</button>
```

javan:
@adrienpoly, you can rebase your branch with upstream master to fix these CI errors: 

Screen Shot 2019-06-11 at 11 20 13 AM

They were resolved in https://github.com/stimulusjs/stimulus/commit/a0fded1bdc6e837e5c0c08465a425d68439cd17f.

javan:
Hi all, quick update:

We've landed on a final syntax for the Values and Classes APIs, and have updated the PR description accordingly. Note the new *Unifying target attributes* section near the end too. We plan to release these changes soon as version 2.0.

You can try the latest build by pointing your `package.json` to:
```json
"stimulus": "https://github.com/stimulusjs/dev-builds/archive/b8cc8c4/stimulus.tar.gz"
```

✌️

walski:
Hey, is there any rough ETA for 2.0 that you can communicate? Thank you so much for all the great work!

dancallaghan:
I've been trying out the new API and I really like it!

It might be good to document "unsetting" values, I had to dig through the source to find that you need to set a value to `undefined` to unset it. At first I wrongly assumed that setting a value to `null` would have the same effect.

dancallaghan:
Also, it would be nice to be able to specify a default value in the value definition.

Just a spitball example:

```javascript
static values = {
  url: String,
  refreshInterval: [Number, 300],
  loadOnConnect: [Boolean, true]
}
```

I understand you could achieve the same thing with a getter method with a fallback in the controller but it seems like the Values API aims to reduce this kind of boilerplate code.

panda-madness:
Is there any update on when this might get merged (and tagged)?

ansonhoyt:
@panda-madness, I'm interested in an official update too, but I've seen a couple hints:
1. @javan [tweeted](https://twitter.com/javan/status/1159898630469378048) "soon" on Aug 9, 2019
2. @dhh [tweeted](https://twitter.com/dhh/status/1178306868667568135) "2020" on Sep 29, 2019

So, we're in the 2020 launch window, but I'm certainly hopeful we don't have 11 months to wait 😄

I *emotionally* feel we're getting teased a bit, but *rationally* understand Basecamp has been shipping lots of new features which would be keeping them busy. Also, they like shipping battle-tested things extracted from their production apps, so that naturally introduces a lag before we get to see it.

jaredcwhite:
@panda-madness @ansonhoyt I third the motion, and as somebody who's using Stimulus on numerous projects now (some in production, some in beta), the lack of visibility into the release process is concerning to me. I don't want to be a bad open source user (_gimme more, gimme it now!_ 😤), but if the process were more in the open, I could maybe help test something or fix a bug or whatnot. Sure I could check out this PR right now and mess around, but I don't really know what I'm doing in there or what is considered ready for release or still a problem.

So how about it Stimulus folks…what can we do to help get this PR done and merged? 😃

dixpac:
Just to chip-in here to say that we are using this branch in production for about month or so, and we didn't see any problems. New changes proved to be really helpful 👍

adrienpoly:
Same here I have been using v2 alpha in production (medium size app) with absolutely no issues

leastbad:
@adrienpoly where did you obtain v2 alpha?

kaspth:
@leastbad posted earlier in the thread 😉
https://github.com/stimulusjs/stimulus/pull/202#issuecomment-520001308

leastbad:
I just want to say that I love this new values API - it's brilliant. The changed callbacks are an unexpected surprise that will make Stimulus even more powerful when wrapping legacy components. I even support switching how targets are defined, even though it will mean multiple attributes on an element that happens to belong to two controllers.

I would love to see @dancallaghan's proposed "spitball" syntax for optionally changing default values.

For these typed values, what is the syntax for the array and object types? For arrays, I assume it'll be comma-separated... will it allow for and filter out spaces?

For object types, what is the data syntax? Will it support objects within objects?

For both arrays and objects, will the elements and values be themselves typed? If I have an array of "5, 5, 5" will these show up as [5, 5, 5] or ["5", "5", "5"]?

If does calling a value setter fire the value changed callback? If not, why?

This might be scope-defying, but could it make sense for a controller to trigger a DOM event when a value setter is called? `identifier:value:changed` with a helpful set of values passed to the details: currentValue, newValue, dataType, external (boolean, setter called by attribute or internal code). This would significantly improve cross-controller communication opportunities.

Final thought for the moment: it would sure be swell if every controller instance could set a read-only, unique GUID value on itself before `initialize()` is called, to make it easier to manage scenarios where you have multiple instances of a given controller on a page.

panda-madness:
I imagine the Basecamp guys have something cooked up in the (hopefully) near future that is at odds with this API, otherwise I'm not sure why this wonderful PR would be seemingly abandoned. I've also tested this PR and loved it, but it feels a little brittle to rely on an API without at least some sort of confirmation of future viability.

leastbad:
@panda-madness @tabuna that's legit, even though the PR came from @sstephenson and they seem to be updating it semi-regularly - admittedly not recently.

Therefore, you should be thrilled that @kaspth commented less than a day ago. I don't think that you're going to get a press release from Basecamp on the matter, so for me this is good enough. ;)

tabuna:
I disagree with you @leastbad . 
A proposal is only a proposal and no more.

We cannot take this to public space. You can’t say to people: "do not look at the official documentation, but look at this thread and its description in the comments." 
We would also like to use other packages that depend on the first version of the stimulus.

leastbad:
@tabuna this isn't described as a proposal. @sstephenson, Stimulus project lead, literally started this PR by saying "We plan to ship them together in the upcoming Stimulus 2.0 release." It is quite literally a heads-up that this will be part of the library in v2, and that's great news.

There's actually nothing here which breaks compatibility with v1. The v1 targets syntax will still work fine, albeit with deprecation warnings in development. This should be a drop-in replacement and other packages shouldn't notice a difference.

That said, you don't need anyone to tell you that if you're worried about using an alpha build in production... don't use it. Nobody is trying to force this on you.

tomekr:
Is there an updated dev-build at this point or should we still be using the link from https://github.com/stimulusjs/stimulus/pull/202#issuecomment-520001308?

kaspth:
@tomekr still the very same, yeah ✌️

leastbad:
I've been using the alpha build of v2 for about a month now, and I love the values API so much.

I do strongly urge the Basecamp folks to consider adopting @dancallaghan's proposed syntax for supporting default values:

> ```js
> static values = {
>   url: String,
>   refreshInterval: [Number, 300],
>   loadOnConnect: [Boolean, true]
> }
> ```

Is there any chance of getting this into the v2 release, @sstephenson @javan @kaspth ?

rik:
To ease working with CSS Utility frameworks (such as [Tailwind](https://tailwindcss.com)), I'm proposing adding a `this.[name]Classes` that would be equivalent to `this.[name]Class.split(' ')`. And would be easy to use with the spread operator to add and remove classes
```html
> ### Class properties
> The Classes API adds a static `classes` array on controllers. As with [targets](https://stimulusjs.org/reference/targets#definitions), Stimulus automatically adds properties for each class listed in the array:
> 
> ```js
> // clipboard_controller.js
> export default class extends Controller {
>   static classes = [ "success", "supported" ]
> 
>   initialize() {
>     if (/* ... */) {
>       this.element.classList.add(this.supportedClass)
>     }
>   }
> 
>   copy() {
>     // ...
>     this.element.classList.add(...this.successClasses)
>   }
> }
> ```
> 
> Kind | Property name | Effect
> --- | --- | ---
> Getter | `this.[name]Class` | Reads the `data-[identifier]-[name]-class` attribute
> Existential | `this.has[Name]Class` | Tests whether the `data-[identifier]-[name]-class` attribute is present

MatheusRich:
Hi folks! I'm using Stimulus with Cleave.js for masks. On the current build (1.1.1) it works like a charm. I'm trying to update to the 2.0-alpha version, but it seems to not work in the same way.

I've created this code pen to examplefy it. Here's an input with a credit card mask. It seems that Stimulus actions run **before** the mask.

https://codepen.io/matheusrich/pen/ZEbXGRB

If you change the library to 1.1.1 (go to JS settings and change `https://srv-file6.gofile.io/download/cVqq34/stimulus.umd.js` to `https://unpkg.com/stimulus`) this does not happen. The action runs `after` the mask is applied.

Any help on this?

leastbad:
@MatheusRich this is really not a great place to ask for this kind of support. Please consider opening an issue or posting on the Stimulus Discourse.

For what it's worth, I checked out your pen and I'm getting `Uncaught ReferenceError: Stimulus is not defined` regardless of which version of the library is installed.

panda-madness:
I legitimately had a dream the other day that Stimulus 2.0 was released. It was glorious, DHH was spraying champagne over a raving crowd and rapping a song about something involving email and pixels.

P.S. Pls no ban.

dhh:
Haha, lol. We really aren't that far off. It's just that there's a lot of CRUNCH on the plate with the release of HEY.

jaredcwhite:
Wait a minute…why is CRUNCH in all caps. Could it be… a new code name?? 😆

Just teasing. Good luck on the impending launch!

panda-madness:
It looks like this API is all but confirmed, if [HEY source code](https://twitter.com/verymadpanda/status/1272816178084315136) is of any indication.

P.S. Grantz on the launch.

leastbad:
@panda-madness This API was confirmed when the guy who makes the library made the PR. 🤷 

If you felt there was any ambiguity on the subject, were you not convinced by repeated comments from Basecamp folks that this is coming?

I'm just very much hoping that there's still time and opportunity for [default values](https://github.com/stimulusjs/stimulus/pull/202#issuecomment-615143760) to make it in before v2 officially launches.

panda-madness:
@leastbad there is no need to be toxic about it.

Scrolling through this thread I cannot see a solid confirmation from any of the Basecamp guys, I asked a [direct question](https://github.com/stimulusjs/stimulus/pull/202#issuecomment-571995108) some time ago that was left unanswered. Hence my perceived ambiguity about this PR's future.

panda-madness:
@sstephenson In the current [alpha build](https://github.com/stimulusjs/stimulus/pull/202#issuecomment-520001308) the lifecycle method invokation order seems to be `initialize -> {valueChanged callbacks} -> connect`. Is that intentional? Intuitively I would have thought that `valueChanged` callbacks would run after `connect`.

msp-matt:
Is there any timeline on when this might get released?

rupasix:
https://twitter.com/dhh/status/1291114354482524160
Maybe we're actually waiting for something different than this PR...

leastbad:
BlaringIdolizedIchthyosaurs-size_restricted

Sploosh

nfilzi:
OMG

PR #206, opened 2018-11-12
Fix image src

wjessop:
👋


javan:
Thanks, Will!

Unfortunately, fixing the image path here will break it on https://stimulusjs.org/handbook/building-something-real.

screen shot 2018-11-13 at 4 52 06 pm

We'll have to do some massaging to get it right in both places.

PR #213, opened 2018-12-17
fix data map documentation

boschr:
Should use the identifier with lowerCamelCase.

### DIFF
--- a/docs/handbook/05_working_with_external_resources.md
+++ b/docs/handbook/05_working_with_external_resources.md
@@ -18,7 +18,7 @@ Begin by sketching the inbox in `public/index.html`:
 
 ```html
 <div data-controller="content-loader"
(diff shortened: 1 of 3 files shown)


sstephenson:
Thanks for the pull request. The case convention for HTML attribute names is kebab-case—see the [Data Maps reference page](https://github.com/stimulusjs/stimulus/blob/master/docs/reference/data_maps.md#naming-conventions) for details.

boschr:
@sstephenson i know, but the `-[reference]-` part (component name) needs to be lowerCamelCase, if your component name consists of multiple words.

sstephenson:
[Controller identifiers are always in kebab-case](https://stimulusjs.org/reference/controllers#naming-conventions).

PR #217, opened 2019-01-03
Add default value param to `get` method of Data Map API

luizbills:
To make possible add optional data attributes.

```js

  initialize() {
    // returns 1 if data-controller-index not exists
    const index = this.data.get('index', 1) 
    this.foo(index);
  }
}
```


javan:
The *Data Map API* is going to be replaced by a new *Values API* soon (see https://github.com/stimulusjs/stimulus/pull/202), which has support for default values. So, I'm going to close this PR for now, but thank you so much for the contribution!

PR #225, opened 2019-01-31
Document keeping controllers out of packs dir

woahdae:
Previously, the documentation for installation had an example implying a directory structure of `app/javascript/packs/controllers`. However, the documentation for rails/webpacker states that packs is a special directory that should only contain entrypoints to be included in HTML, ex. application.js.

This changes the installation example to imply a directory structure of `app/javascript/controllers`, and explicitly gives that structure as context to the example.

### DIFF
--- a/docs/handbook/06_installing_stimulus.md
+++ b/docs/handbook/06_installing_stimulus.md
@@ -14,7 +14,7 @@ To install Stimulus in your application, add the [`stimulus` npm package](https:
 
 Stimulus integrates with the [webpack](https://webpack.js.org/) asset packager to automatically load controller files from a folder in your app.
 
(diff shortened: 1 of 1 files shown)


woahdae:
Pull request inspired by discussion [on the discourse list](https://discourse.stimulusjs.org/t/new-to-stimulus-tiny-controller-directory-placement-question/643/2)

javan:
Thanks for bringing this up!

The Stimulus documentation tries to steer clear of framework (e.g. Rails) specifics, and its install guide includes a section for webpack, but not webpack<em><strong>er</strong></em>, which has its own set of conventions. If you're using Rails + Webpacker, I suggest using the [baked in Stimulus installer](https://github.com/rails/webpacker#stimulus). It creates a `controllers/` directory adjacent to `packs/`, avoiding the issue/question you ran into here.

So, we're going to pass on this change for now, and will consider updating the docs to nudge people using Rails in the right direction. Thanks again, and sorry for the confusion.

PR #231, opened 2019-03-19
link to the browser version of the closest polyfill

### LINKED ISSUE #230: @stimulus/polyfills : iOS9 Unexpected keyword 'const'. Const declarations are not supported in strict mode.
I had some compatibility issues with an older iOS device (9.x) when using Stimulus

The console shows the following error:
```
Unexpected keyword 'const'. Const declarations are not supported in strict mode.
```

The code related to this error is the following
```js
function polyfill(window) {
  const ElementPrototype = window.Element.prototype;

  if (typeof ElementPrototype.matches !== 'function') {
    ElementPrototype.matches = ElementPrototype.msMatchesSelector || ElementPrototype.mozMatchesSelector || ElementPrototype.webkitMatchesSelector || function matches(selector) {
      const elements = (element.document || element.ownerDocument).querySelectorAll(selector);
      let element = this;
      let index = 0;

      while (elements[index] && elements[index] !== element) {
        ++index;
      }

      return Boolean(elements[index]);
    };
  }

  if (typeof ElementPrototype.closest !== 'function') {
    ElementPrototype.closest = function closest(selector) {
      let element = this;

      while (element && element.nodeType === 1) {
        if (element.matches(selector)) {
          return element;
        }

        element = element.parentNode;
      }

      return null;
    };
  }
}
```

This is the `element.closest` polyfill that is part of the `@stimulus/polyfills` package

I am in a Rails environment
I tried to change my babel setting to something like that 
```
{
  "presets": [
    [
      "env",
      {
        "modules": false,
        "targets": {
          "browsers": ["last 2 versions", "safari >= 7"],
          "uglify": true
        },
        "useBuiltIns": true
      }
    ]
  ]
...
}
```

but it didn't change anything. (I think webpacker default setting will not transpile node modules it expects it to have already been transpiled)

I found two solutions:
1) `import "babel-polyfill";` instead of `@stimulus/polyfills`
2) change the `@stimulus/polyfills` to import the browser version of element.closest polyfills

```js
//polyfills/index.js
...
import "element-closest/browser.js";
```

adrienpoly:
Get the browser version of the closest polyfill
fix #230


adrienpoly:
Ok actually I realized too late that this was fixed in 1.1.1 by specifying a 2.x version of the polyfill. 
This PR therefore only bumps the closest polyfill version to 3.x and loads the correct polyfill browser version.
I didn't see any clear benefit between 3.0.1 and 2.x series
so unless you want to keep this version bump you can close this PR

javan:
Thanks, and sorry for the confusion. It was a surprising change in 3.x, which is why we just locked to 2.x for now. Will go this route if/when we bump up to 3.

PR #232, opened 2019-03-22
Introduce once and passive events to actions

adrienpoly:
As per the discussion in this [thread](https://discourse.stimulusjs.org/t/passive-event-listeners/709), this pull request introduces two new options for events attached to an action 

- **once**
- **passive**
```js
addEventListener('click', action, {passive: false, once: false})
```

The proposed API is the following
```html
<!--passive: true-->
<div data-action="touchmove.passive->controller#method"></div>
<div data-action="touchmove@window.passive->controller#method"></div>

<!--once: true-->
<button data-action="click.once->controller#vote"></button>
<div data-action="keydown@window.once->controller#delete"></div>
```

### Defaults events
**Defaults events with options are not supported**

an action like this 

```html
<button data-action="controller#method">
```
 must be written like this to pass an option:

```html
<button data-action="click.once->controller#method">
```

**considerations:**
I see two possible syntaxes to support default event with options

1)
```html
<button data-action="once.controller#method">
```
 
2)
```html
<button data-action="controller#method.once">
// and this would be generalized 
<div data-action="touchmove->controller#method.passive"></div>
<div data-action="touchmove@window->controller#method.passive"></div>
```
open to discussion....

### Tests
I added a new test case for those events in `src/tests/cases/event_options_tests.ts`. The tests for `once` events are ok IMHO but I am having a hard time testing the `passive` event

When I test in a real app the passive event I can get it to work but when I log as below an action in my test suite I always get non-passive events.

```js
logPassive(event: Event) {
    event.preventDefault()
    if (event.defaultPrevented) {
     //this is not a passive event
      this.recordAction("passive", event, false)
    } else {
      //this is a passive event
      this.recordAction("passive", event, true)
    }
  }
```
Will keep investigating, maybe I am missing something here. Could it be that the chrome headless version does not support passive event???

### Example
once a **dev-build** is available (not sure how new build are trigged) here is a test example
https://glitch.com/edit/#!/stimulus-event-options?path=src/controllers/options_controller.js:15:23

### Still to do…
- [x] fix test for passive events
- [x] do we need to handle default event with a simpler API?
- [x] key/value syntax to support all event listener options?
- [x] Polyfill for ie11? potential [solution](https://github.com/github/eventlistener-polyfill) but it does not support `handleEvent` 
- [x] documentation


kaspermeyer:
Nice work! I was wondering if it would be possible to expand the API a bit to ensure that more than one event listener option can be added per action and provide support for other options such as [`bubbles`](https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles), [`cancelable`](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable) and [`composed`](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed)?

I'm not sure if it should be a whitelisted set of properties or if any arbitrary property should be accepted. Also, allowing custom values per option makes sense for properties like `bubbles` which are not `true` by default.

I considered using a quoteless JSON format, but it doesn't feel very easy on the eyes:
```html
<div data="event{bubbles:false,once:true}->controller#action">
```

Perhaps it could be along the lines of:
```html
<div data="event{bubbles:false|once:true}->controller#action">
```

adrienpoly:
Thanks

> I was wondering if it would be possible to expand the API a bit to ensure that more than one event listener option can be added per action and provide support for other options such as [`bubbles`](https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles), [`cancelable`](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable) and [`composed`](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed)?

I am not sure to really understand, maybe I am missing somehting here but `passive` and `once` are direct options of the [eventListener ](https://developer.mozilla.org/fr/docs/Web/API/EventTarget/addEventListener)and the eventListener is defined[ inside Stimulus with a a current hard coded value of `false`](https://github.com/stimulusjs/stimulus/blob/master/packages/%40stimulus/core/src/event_listener.ts#L15) for both. So it would make sens to pass them as an option.

For `Bubbles`, `cancelable` they are options of the event by itself not the listener, Stimulus don't create events, it mostly listen to events and dispatch actions. So I am not really sure of what would be the use of defining bubbles/cancelable at the action level?

kaspermeyer:
You're totally right, I'm sorry about the confusion! I got the options of the event listener and the events themselves mixed up 😕 Thanks for taking the time to sort it out.

javan:
There are still good reasons for providing a key/value syntax to support all event listener options. For example, `touchmove` is `passive: true` by default in Chrome and you may want to negate that.

adrienpoly:
ok thanks, both for your inputs
how about `.not_passive` to explicitly set it to false and to keep a short syntax?

Does it makes sens to have `not_once`?
Does it makes sens to be able to combine them `.passive.once`?

kaspermeyer:
I'm tempted to suggest antonyms for the options like **_passive / active_** and **_once / always_**, but it might be too ambiguous since **_passive / once_** represents actual option names and **_active / always_** does not.

I know Sam mentioned that [he liked the dot notation](https://discourse.stimulusjs.org/t/passive-event-listeners/709/2), but I'm wondering if it's too close to the target descriptor? The JSON notation `event{passive:false,once:true}->controller#action` (or a variation of it) doesn't conflict with existing descriptor patterns, it solves the problem of negating values without ambiguity, but it's a bit verbose. What do you think?

sstephenson:
Here’s my syntax suggestion:

```html
<!--passive: true-->
<div data-action="touchmove[passive]->controller#method"></div>
<div data-action="touchmove[passive]@window->controller#method"></div>

<!--once: true-->
<button data-action="click[once]->controller#vote"></button>
<div data-action="keydown[once]@window->controller#delete"></div>
<button data-action="[once]controller#vote"></button>

<!--passive: false-->
<div data-action="touchmove[!passive]->controller#method"></div>

<!--passive: false, once: true-->
<div data-action="touchmove[!passive][once]->controller#method"></div>
```

* Options are boolean tags attached to the right-hand side of the event name (_before_ the event target specifier, if one is present).
* An option with a true value is written as `[option]`; an option with a false value, as `[!option]`.
* If the action observes a default event, options can be attached to the left-hand side of the controller identifier: `[option]controller#method`.
* Square brackets are visually calmer than curly braces, but they don’t look like function arguments, as parentheses might.

sstephenson:
Thinking through my proposal a bit more, there’d really be only four new tokens to match: `[passive]`, `[!passive]`, `[once]`, and `[capture]`. (I can’t see any reason you’d need to specify `once: false` or `capture: false` since they’re always always the default, regardless of the event type.)

ngan:
How about: `touchmove(once,!passive)->controller#method`? Using parens makes it look like arguments.

sstephenson:
In this case, I think parens are a bad idea _because_ they look like arguments.

A common question is [how to pass arguments to action methods](https://github.com/stimulusjs/stimulus/issues/64). I can see how someone might confuse a parenthesized option syntax with arguments being passed to the action itself.

adrienpoly:
After some additional thought on this, I think within the various current proposals we have too much syntax variants for this feature:
- standard syntax `<div data-action="touchmove[passive]->controller#method"></div>`
- default events `<button data-action="[once]controller#vote"></button>`
- global events `<div data-action="keydown[once]@window->controller#delete"></div>`

It makes it hard to remenber where and how to put the options

### Suggestion use the same syntax as the addEventListener

At the end what we are doing is specifying all of the arguments of the event listener.  My idea  is to keep the order and syntax as close as possible to the àddEventListener``


Screen Shot 2019-03-29 at 8 41 48 PM

**Keeping the same order would translate to such a syntax**

Screen Shot 2019-03-29 at 8 41 57 PM

Moving the options to the end simplifies the syntaxes, options always at the end whether there is a global event or a default event and we keep the same order so it is easy to remember.

### Bracket syntax

```html
<!--passive: true-->
<div data-action="touchmove->controller#method[passive]"></div>
<div data-action="touchmove@window->controller#method[passive]"></div>

<!--once: true-->
<button data-action="click->controller#vote[once]"></button>
<button data-action="controller#vote[once]"></button>

<!--passive: false-->
<div data-action="touchmove->controller#method[!passive]"></div>

<!--passive: false, once: true-->
<div data-action="touchmove->controller#method[!passive][once]"></div>
``` 

### Dot syntax

```html
<!--passive: true-->
<div data-action="touchmove->controller#method.passive"></div>
<div data-action="touchmove@window->controller#method.passive"></div>

<!--once: true-->
<button data-action="click->controller#vote.once"></button>
<button data-action="controller#vote.once"></button>

<!--passive: false-->
<div data-action="touchmove->controller#method!passive"></div>

<!--passive: false, once: true-->
<div data-action="touchmove->controller#method!passive.once"></div>
<div data-action="touchmove->controller#method.once!passive"></div>
```

kaspermeyer:
I like the idea of moving the suffix to the action and reduce the number of variations. That might have been why I initially mixed up the event and listener options: they were declared at the event part of the descriptor instead of at the action (which plays the role of a listener).

sstephenson:
Fair point about matching the argument order of `addEventListener`.

If we put options on the right, I don’t think we should use brackets or dots. They both look too much like property accessors. A colon works nicely (and evokes CSS pseudo-selector syntax):

```html
<!--passive: true-->
<div data-action="touchmove->controller#method:passive"></div>
<div data-action="touchmove@window->controller#method:passive"></div>

<!--once: true-->
<button data-action="click->controller#vote:once"></button>
<button data-action="controller#vote:once"></button>

<!--passive: false-->
<div data-action="touchmove->controller#method:passive"></div>

<!--passive: false, once: true-->
<div data-action="touchmove->controller#method:!passive:once"></div>
```

adrienpoly:
ok thanks, @sstephenson looks like we all together came up to a nice plan. Will update the code accordingly

conormuirhead:
Just wanted to pop in and say that it's neat to see how y'all worked through this together. 👏 Great job, and thanks for doing this kind of work out in the open so others like me can benefit!

adrienpoly:
I have updated the code to reflect this new syntax.

several comments:
- Took me some time to figure it out but I had to differentiate the key for the `eventListenerMap` on both the event name and the option. Otherwise when chaining multiple actions with the same event all subsequent action would have the same option as the first one

adrienpoly:
I have updated the code to reflect this new syntax.

allowed tokens are `"passive", "!passive", "once", "capture"`

#### Several comments:

1. **eventListenerMap**: Took me some time to figure it out but I had to differentiate the key for the `eventListenerMap` on both the event name and the option. Otherwise, when chaining multiple actions with the same event name all subsequent actions would have the same option as the first one

A typical example would be a button to vote, first click increase a counter with stopPropagation and subsequent clicks display an alert message
```html
<button data-action="controller#vote:once controller#alert"></button>
```

To do so I modified the key of the  `eventListenerMap` to being a combination of the event name and options ```JSON.stringify({ eventName, ...eventOptions})```. All test pass (locally!) but this is probably a sensitive part that needs to be [reviewed carefully](https://github.com/adrienpoly/stimulus/blob/passive_events/packages/%40stimulus/core/src/dispatcher.ts#L59-L67)

2. **Testing passive events**: initially I wasn't able to test passive events. I finally discovered [those lines of code](https://github.com/adrienpoly/stimulus/blob/f01a72b038f314d18b0547c3bc363034e66d1b34/packages/%40stimulus/test/src/dom_test_case.ts#L31-L34) that were preventing testing whether and action is passive or not. Those lines probably have a good reason to be here so for now I created a duplicate of this function just to test the passive events.

3. **Polyfill**: all test pass locally but fails in the CI on some browser. I tried adding [this polyfill](https://github.com/github/eventlistener-polyfill) but while it improves the compatibility it does still fails a lot. Does anyone have a recommended alternative to polyfill once and passive option?

adrienpoly:
We can get all tests green by using the Github Polyfill AND explicitly passing the handleEvent function to the `addEventListener`

javan:
> We can get all tests green by using the Github Polyfill AND explicitly passing the handleEvent function to the `addEventListener`

I'd rather not make code changes to work around an incomplete polyfill. Perhaps we can instead update the polyfill to support the [`EventListener` interface](https://developer.mozilla.org/en-US/docs/Web/API/EventListener)?

adrienpoly:
> I'd rather not make code changes to work around an incomplete polyfill. Perhaps we can instead update the polyfill to support the [`EventListener` interface](https://developer.mozilla.org/en-US/docs/Web/API/EventListener)?

Ok @javan I ll try to open a PR on the respective repo with the support for `EventListner` I think I see how this could be added.  Not sure however the PR would be merged anytime soon.

javan:
I opened https://github.com/github/eventlistener-polyfill/pull/13 to add `EventListener` interface support. Unclear if that project being maintained though.

adrienpoly:
Thanks @javan I was caught up on some other projects and didn't have time yet to look into it. That is great.

Do you want that I update this branch with the polyfill code directly into `@stimulus/polyfill` just to confirm that with this solution we pass all tests on the CI?

We can then decide whether we provide our own polyfill or wait for the PR to be merged and published

javan:
No need to add the code directly. You can apply this change to test with my fork:
<details>
<summary>View diff</summary>

```diff
diff --git a/packages/@stimulus/core/src/event_listener.ts b/packages/@stimulus/core/src/event_listener.ts
index 018e2bf..3254fe2 100644
--- a/packages/@stimulus/core/src/event_listener.ts
+++ b/packages/@stimulus/core/src/event_listener.ts
@@ -14,11 +14,11 @@ export class EventListener implements EventListenerObject {
   }
 
   connect() {
-    this.eventTarget.addEventListener(this.eventName, this.handleEvent, this.eventOptions)
+    this.eventTarget.addEventListener(this.eventName, this, this.eventOptions)
   }
 
   disconnect() {
-    this.eventTarget.removeEventListener(this.eventName, this.handleEvent, false)
+    this.eventTarget.removeEventListener(this.eventName, this, false)
   }
 
   // Binding observer delegate
@@ -33,7 +33,7 @@ export class EventListener implements EventListenerObject {
     this.unorderedBindings.delete(binding)
   }
 
-  handleEvent = (event: Event) => {
+  handleEvent(event: Event) {
     const extendedEvent = extendEvent(event)
     for (const binding of this.bindings) {
       if (extendedEvent.immediatePropagationStopped) {
diff --git a/packages/@stimulus/polyfills/index.js b/packages/@stimulus/polyfills/index.js
index 0280022..e3b67d6 100644
--- a/packages/@stimulus/polyfills/index.js
+++ b/packages/@stimulus/polyfills/index.js
@@ -7,4 +7,4 @@ import "core-js/fn/promise"
 import "core-js/fn/set"
 import "element-closest"
 import "mutation-observer-inner-html-shim"
-import "eventlistener-polyfill"
+import "eventlistener-polyfill/src"
diff --git a/packages/@stimulus/polyfills/package.json b/packages/@stimulus/polyfills/package.json
index b6459f5..0d6e819 100644
--- a/packages/@stimulus/polyfills/package.json
+++ b/packages/@stimulus/polyfills/package.json
@@ -13,7 +13,7 @@
     "core-js": "^2.5.3",
     "element-closest": "^2.0.2",
     "mutation-observer-inner-html-shim": "^1.0.0",
-    "eventlistener-polyfill": "^1.0.4"
+    "eventlistener-polyfill": "javan/eventlistener-polyfill#event-listener-interface"
   },
   "publishConfig": {
     "access": "public"
```

</details>

adrienpoly:
All green 🚀

packages/@stimulus/core/src/event_listener.ts
```diff
@@ -3,16 +3,18 @@ import { Binding } from "./binding"
...
     this.eventTarget = eventTarget
     this.eventName = eventName
-    this.unorderedBindings = new Set
+    this.eventOptions = this.convertEventOptions(eventOptions)
+    this.unorderedBindings = new Set()
   }
 
   connect() {
-    this.eventTarget.addEventListener(this.eventName, this, false)
+    this.eventTarget.addEventListener(this.eventName, this, this.eventOptions)
```
sstephenson: We’ll also need to pass `this.eventOptions` as the third argument to `removeEventListener()` in `disconnect()` below.

docs/reference/actions.md
```diff
@@ -36,11 +36,12 @@ An action is a connection between:
 
 ## Descriptors
 
-The `data-action` value `click->gallery#next` is called an _action descriptor_. In this descriptor:
+The `data-action` value `click->gallery#next:once` is called an _action descriptor_. In this descriptor:
```
sstephenson: Let’s revert the change to this example and the addition to the list below. Most actions won’t have any options and I think it’s a little overwhelming to see it up front. (The Options section below is good 👍)

packages/@stimulus/core/src/action_descriptor.ts
```diff
@@ -1,22 +1,25 @@
...
+  eventOptions: AddEventListenerOptions
   eventName: string
   identifier: string
   methodName: string
 }
 
-// capture nos.:            12   23 4               43   1 5   56 7  76
-const descriptorPattern = /^((.+?)(@(window|document))?->)?(.+?)(#(.+))?$/
+// capture nos.:            12   23 4               43   1 5   56 7      768 9  98
+const descriptorPattern = /^((.+?)(@(window|document))?->)?(.+?)(#([^:]+?))(:(.+))?$/;
```
sstephenson: ✂️ trailing semicolon

packages/@stimulus/core/src/action_descriptor.ts
```diff
@@ -1,22 +1,25 @@
...
-    eventName:   matches[2],
-    identifier:  matches[5],
-    methodName:  matches[7]
-  }
+    eventTarget:  parseEventTarget(matches[4]),
+    eventName:    matches[2],
+    identifier:   matches[5],
+    methodName:   matches[7],
+    eventOptions: parseEventOptionTokens(matches[9]),
+  };
```
sstephenson: ✂️ trailing semicolon

packages/@stimulus/core/src/dispatcher.ts
```diff
@@ -52,22 +52,22 @@ export class Dispatcher implements BindingObserverDelegate {
...
     const eventListenerMap = this.fetchEventListenerMapForEventTarget(eventTarget)
-    let eventListener = eventListenerMap.get(eventName)
+    let eventListener = eventListenerMap.get(JSON.stringify({ eventName, ...eventOptions}))
     if (!eventListener) {
-      eventListener = this.createEventListener(eventTarget, eventName)
-      eventListenerMap.set(eventName, eventListener)
-    }
+      eventListener = this.createEventListener(eventTarget, eventName, eventOptions)
+      eventListenerMap.set(JSON.stringify({ eventName, ...eventOptions }), eventListener)
+    }
```
sstephenson: ✂️ trailing whitespace

packages/@stimulus/core/src/event_listener.ts
```diff
@@ -3,16 +3,18 @@ import { Binding } from "./binding"
 export class EventListener implements EventListenerObject {
   readonly eventTarget: EventTarget
   readonly eventName: string
+  readonly eventOptions: AddEventListenerOptions | false
   private unorderedBindings: Set<Binding>
 
-  constructor(eventTarget: EventTarget, eventName: string) {
+  constructor( eventTarget: EventTarget, eventName: string, eventOptions: AddEventListenerOptions) {
```
sstephenson: ✂️ whitespace after the open paren

packages/@stimulus/core/src/tests/cases/event_options_tests.ts
```diff
@@ -0,0 +1,115 @@
...
+      <button></button>
+    </div>
+    <div id="outside"></div>
+  `
+  async "test different syntaxes for once action"() {
+    this.actionValue = "click->c#log:once d#log2:once c#log3:once"
+    
+    await this.nextFrame
+    await this.triggerEventPassivable(this.buttonElement, "click");
+    await this.triggerEventPassivable(this.buttonElement, "click");
```
sstephenson: ✂️ trailing semicolons

packages/@stimulus/core/src/tests/cases/event_options_tests.ts
```diff
@@ -0,0 +1,115 @@
...
+      { name: "log3", identifier: "c", eventType: "click", currentTarget: this.buttonElement },
+    )
+  }
+
+  async "test mix once and standard actions"() {
+    this.actionValue = "c#log:once d#log2 c#log3";
+    
+    await this.nextFrame
+    await this.triggerEventPassivable(this.buttonElement, "click");
+    await this.triggerEventPassivable(this.buttonElement, "click");
```
sstephenson: ✂️ trailing semicolons

packages/@stimulus/core/src/tests/cases/event_options_tests.ts
```diff
@@ -0,0 +1,115 @@
...
+      { name: "log2", identifier: "d", eventType: "click", currentTarget: this.buttonElement },
+      { name: "log3", identifier: "c", eventType: "click", currentTarget: this.buttonElement },
+    )
+  }
+
+  async "test stop propagation with once"() {
+    this.actionValue = "c#stop:once c#log"
+
+    await this.nextFrame
+    await this.triggerEventPassivable(this.buttonElement, "click");
```
sstephenson: ✂️ trailing semicolon

packages/@stimulus/core/src/tests/cases/event_options_tests.ts
```diff
@@ -0,0 +1,115 @@
...
+      { name: "log", identifier: "c", eventType: "click", currentTarget: this.buttonElement },
+    )
+  }
+
+  async "test global once actions"() {
+    this.actionValue = "keydown@window->c#log:once"
+    
+    await this.nextFrame
+    await this.triggerEventPassivable("#outside", "keydown");
+    await this.triggerEventPassivable("#outside", "keydown");
```
sstephenson: ✂️ trailing semicolons

packages/@stimulus/core/src/tests/cases/event_options_tests.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    this.assertActions({ name: "log", eventType: "keydown" })
+  }
+
+  async "test default passive action"() {
+    this.actionValue = "scroll->c#logPassive:passive"
+    await this.nextFrame
+
+    await this.triggerEventPassivable(this.buttonElement, "scroll");
```
sstephenson: ✂️ trailing semicolon

packages/@stimulus/core/src/tests/cases/event_options_tests.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  get element() {
+    return this.findElement("div")
+  }
+
+  get buttonElement() {
+    return this.findElement("button")
+  }
+}
```
sstephenson: missing newline at the end of the file

packages/@stimulus/test/src/dom_test_case.ts
```diff
@@ -38,6 +38,16 @@ export class DOMTestCase extends TestCase {
     return event
   }
 
+  async triggerEventPassivable(selectorOrTarget: string | EventTarget, type: string, bubbles: boolean = true) {
```
sstephenson: Is this method necessary just to avoid the IE <= 11 `preventDefault()` implementation in `triggerEvent()`? Maybe we can figure out how to condition it away instead. E.g. maybe the third argument to `triggerEvent()` can become an options object.

javan:
@adrienpoly, you can rebase your branch with upstream master to fix these CI errors: 

Screen Shot 2019-06-11 at 11 20 13 AM

They were resolved in https://github.com/stimulusjs/stimulus/commit/a0fded1bdc6e837e5c0c08465a425d68439cd17f.

packages/@stimulus/core/src/action_descriptor.ts
```diff
@@ -27,6 +30,20 @@ function parseEventTarget(eventTargetName: string): EventTarget | undefined {
...
+  const regexBoolean = /^[^!]{1}.+$/; // ! negates a token
+
+  return tokens.reduce(
+    (result, option) => ({
+      ...result,
+      [option.replace(/!/,"")]: regexBoolean.test(option)
+    }),
+    {} as AddEventListenerOptions
+  )
+}
```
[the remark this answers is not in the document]
adrienpoly: I tried using this function but got some errors on IE with the CI 
https://circleci.com/gh/stimulusjs/stimulus/502

For now I reverted to my previous function, I don't have time today to look further into this but your function is much cleaner and simpler. I ll try to find out what is the difference? and update the code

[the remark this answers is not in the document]
adrienpoly: I also added a test just to make sure wrong are silently ignored

adrienpoly:
@sstephenson I have updated the code with all your reviews, there is one comment from Javan that I have not yet added as it is causing some errors in the CI on IE that I don't fully understand right now. I will look into this a bit later

I [added a test](https://github.com/stimulusjs/stimulus/blob/d76094cdbecf1a24dcfcce528e21ad0c689841c0/packages/%40stimulus/core/src/tests/cases/event_options_tests.ts#L69-L85) for your edge case.
The test pass without any change..... So I think we are good unless my test does not fully reflect your case?

packages/@stimulus/core/src/action_descriptor.ts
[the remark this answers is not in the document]
adrienpoly: @javan ok I understood the cause of the failures on the CI, IE11 does not support `startsWith`
I replaced it with a simple regex test `!/^!/.test(token)` and the full function becomes:

```js
function parseEventOptions(eventOptions: string = ""): AddEventListenerOptions {
  return eventOptions.split(":").reduce((options, token) =>
    Object.assign(options, { [token.replace(/^!/, "")]: !/^!/.test(token) })
    , {})
}
```

javan:
Thank you, @adrienpoly!!

PR #237, opened 2019-04-01
@stimulus/polyfills: Polyfill `Reflect.construct`

seanpdoyle:
When calling `blessDefinition` from the `@stimulus/core/definition`
package, the private `extend` function [relies on `Reflect.construct`][call-site],
which is [unsupported in Internet Explorer 11][mdn].

This commit imports `Reflect.construct` from the `core-js` alongside the
other polyfills.

[call-site]: https://github.com/stimulusjs/stimulus/blob/7f58b5597292e88423c3fa9ccd0e6f25ffc83192/packages/%40stimulus/core/src/definition.ts#L27
[mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/construct


javan:
`Reflect.construct` is only used in browsers that support it and there's a fallback implementation for those that don't: https://github.com/stimulusjs/stimulus/blob/7f58b5597292e88423c3fa9ccd0e6f25ffc83192/packages/%40stimulus/core/src/definition.ts#L45-L50

Are you seeing related errors in IE 11?

seanpdoyle:
> Are you seeing related errors in IE 11?

Yup! I discovered the need for a `Reflect.construct` polyfill in our application when troubleshooting in IE11.

I didn't notice the fallback conditional toward the bottom of this file. 

From what I saw while troubleshooting, without the polyfill, the fallback doesn't seem to be effective.

What sort of information would be helpful for me to provide?

javan:
> What sort of information would be helpful for me to provide?

Can you create a bare bones Stimulus app on glitch (or elsewhere) that fails in IE 11?

seanpdoyle:
@javan I've created https://amethyst-piano.glitch.me/ .

To keep things simple, I've included Stimulus through a `<script src="https://unpkg.com...">` tag, along with a `babel-standalone` tag.

I'm not sure how to add the polyfills with `unpkg.com`, since they don't include a `dist` version.

If you don't know of a simple way to consume the polyfills in the same style, I'll do my best to create a repro repo.

javan:
Here's one that includes `@stimulus/polyfills`: https://glitch.com/~atom-theory

I'm not seeing any errors in IE 11:
Screen Shot 2019-04-02 at 5 47 43 PM

seanpdoyle:
Thanks for creating that Glitch @javan!

I attempted my own (https://glitch.com/edit/#!/distinct-cyclone), but had some trouble integrating the polyfills.

> I'm not seeing any errors in IE 11:

That's interesting. I'm not seeing any when visiting that Glitch either.

The application that this issue arose in includes some other polyfills to make use of Custom Elements, so the two sets of polyfills might've clobbered one another or not integrated properly.

I'm going to close this. 

Thank you!

PR #239, opened 2019-04-02
Add UMD global TS declaration

### LINKED ISSUE #238: "export as namespace" in index.d.ts
I want to use Stimulus as a UMD global from a **non-module-based** TypeScript project.

```typescript
// foo_controller.ts
class FooController extends Stimulus.Controller {
    // ...
}
```

The `index.d.ts` file which ships with Stimulus only declares a TypeScript module, not a UMD global as well, which means that you can't use Stimulus in a TS script file.

Please could you add [`export as namespace Stimulus`](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions) to `index.d.ts`?

benjamin-hodgson:
Fixes #238

### DIFF
--- a/packages/stimulus/index.d.ts
+++ b/packages/stimulus/index.d.ts
@@ -1,1 +1,2 @@
 export * from "@stimulus/core"
+export as namespace Stimulus


sstephenson:
Thank you!

PR #264, opened 2019-08-07
Fix for DataMap set() method always returning null

chrisgreen1993:
This was caused by the key parameter being mutated to the formatted version before being passed into get - causing it to never be found


javan:
Thank you!

PR #266, opened 2019-08-12
Optimize finding singular target elements

javan:
Tested (somewhat unscientifically) using the following markup nested several levels deep:
```html
<div data-controller="x" data-target="x.a">
  <span>
    <span data-target="x.b"></span>
    <span data-target="x.b"></span>
  </span>

  <span data-target="x.c"></span>
  <span data-target="x.c"></span>

  <div data-controller="x" data-target="x.a">
    …
  </div>
</div>
```

Before:
```
this.aTarget 0.140ms
this.bTarget 0.230ms
this.cTarget 0.160ms
```

After:
```
this.aTarget 0.035ms
this.bTarget 0.140ms
this.cTarget 0.110ms
```

### DIFF
--- a/packages/@stimulus/core/src/scope.ts
+++ b/packages/@stimulus/core/src/scope.ts
@@ -19,21 +19,24 @@ export class Scope {
   }
 
   findElement(selector: string): Element | undefined {
-    return this.findAllElements(selector)[0]
+    return this.element.matches(selector)
+      ? this.element
+      : this.queryElements(selector).find(this.containsElement)
   }
 
   findAllElements(selector: string): Element[] {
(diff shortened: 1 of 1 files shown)


sstephenson:
The code itself looks fine 👍 I’ll leave it up to you as far as merging, though.

I’d prefer not to make micro-optimizations like this without some sort of corpus we can benchmark against. As you’ve noted, the test case here is unscientific and not really representative of an actual DOM tree, and AFAIK we don’t have any reason to think the current implementation will be a performance problem. We also wouldn’t have any way to know if we were to regress in the future.

PR #271, opened 2019-08-24
Fix links to Glitch

digitalfrost:
Use new Glitch import from git url format

Fixes: https://github.com/stimulusjs/stimulus/issues/270


javan:
Thanks!

PR #281, opened 2019-10-18
Modernize flagship example

tigt:
The old example HTML seems very 4.01 — unless compatibility with old IE is a goal of Stimulus, moving to more idiomatic, accessible, and user-friendly HTML seems like an improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@ How does it work? Sprinkle your HTML with controller, target, and action attribu
 
 ```html
 <div data-controller="hello">
-  <input data-target="hello.name" type="text">
+  <input data-target="hello.name" autocomplete="given-name">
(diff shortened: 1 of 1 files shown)


tigt:
Personally, I would leverage Stimulus even more with the shorthand action notation and better keyboard UX with something like:

```html
<form data-controller="hello" data-action="hello#greet">
  <input data-target="hello.name" autocomplete="given-name">

  <button>Greet</button>

  <output data-target="hello.output"></output>
</form>
```

```js
static targets = [ "name", "output" ]
greet() {
    this.outputTarget.value =
      `Hello, ${this.nameTarget.value}!`
  }
}
```

…but that seemed too extreme to propose.

PR #283, opened 2019-11-21
Update URL TLD to .dev

contributionist:


sstephenson:
That isn’t our domain.

PR #285, opened 2020-01-22
Add Nested Scope to the controllers reference doc

holdenrehg:
Provides a short description and example of how scope behaves for nested controllers.

I was reading through the docs and some examples getting a feel for the framework in general and noticed the nested scope behavior. Then I saw [this discourse post](https://discourse.stimulusjs.org/t/are-recursive-controllers-explicitly-supported/421).

I think having even a brief description in the docs will especially help out devs new to the framework.

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -61,7 +61,7 @@ For example, the `<div>` and `<h1>` below are part of the controller's scope, bu
 
 ## Nested Scopes
 
-When controllers are nested, they're aware of _only_ their own scope and not the scope of any controllers nested within them.
+When nested, each controller is only aware of its own scope excluding the scope of any controllers nested within.
 
 For example, the `#parent` controller below is only aware of the `list.item` targets directly within its scope, but not any targets of the `#child` controller.
 
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -59,6 +59,25 @@ For example, the `<div>` and `<h1>` below are part of the controller's scope, bu
 </main>
 ```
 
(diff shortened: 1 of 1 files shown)


holdenrehg:
@sstephenson this is my first PR to `stimulus`. I read through code of conduct and such, but let me know if I should be following a different process or tagging certain users as reviewers.

docs/reference/controllers.md
```diff
@@ -59,6 +59,25 @@ For example, the `<div>` and `<h1>` below are part of the controller's scope, bu
 </main>
 ```
 
+## Nested Scopes
+
+When controllers are nested, they're aware of _only_ their own scope and not the scope of any controllers nested within them.
```
robzolkos: I'd change to `they are`.   But perhaps `they` should be explicit about the parent.   So `When nested, the parent controller is only aware of their own scope...`

holdenrehg: Ya, the wording was a bit tricky. I guess technically every controller is only aware of its own scope, no matter where it is in the tree.

What do you think about something like the following? Think it's necessary to be explicit about the `parent`?

```
When nested, each controller is only aware of its own scope excluding the scope of any controllers nested within.
```

hms: This just burned me and is important to document sooner rather than later as the current documentation is incorrect:  "When Stimulus connects a controller to an element, that element and all of its children make up the controller's scope."

While the current behavior is understandable and probably even desirable in most cases, there are cases where it's not.  Wondering if there is anyway to have an option to allow a controllers scope to include the Dom of nested controllers?

PR #296, opened 2020-03-15
Improve Naming Conventions documents

JuanitoFatas:
Hi 👋 

This Pull Request improves [naming conventions] doc. Two changes

- Mention snake_case first since it got used more frequently in the docs
- Add `_controller.js` suffix to filenames

Before

スクリーンショット 2020-03-15 13 39 08

After

スクリーンショット 2020-03-17 10 04 19

[naming conventions]: https://stimulusjs.org/reference/controllers#naming-conventions

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -92,7 +92,7 @@ Always use camelCase for method and property names in a controller class.
 
 When an identifier is composed of more than one word, write the words in kebab-case (i.e., by using dashes: `date-picker`, `list-item`).
 
-In filenames, separate multiple words using either dashes or underscores (kebab-case or snake_case: `controllers/date_picker.js`, `controllers/list-item.js`).
+In filenames, separate multiple words using either underscores or dashes (snake_case or kebab-case: `controllers/date_picker_controller.js`, `controllers/list-item-controller.js`).
 
 ## Registration
 


docs/reference/controllers.md
```diff
@@ -92,7 +92,7 @@ Always use camelCase for method and property names in a controller class.
 
 When an identifier is composed of more than one word, write the words in kebab-case (i.e., by using dashes: `date-picker`, `list-item`).
 
-In filenames, separate multiple words using either dashes or underscores (kebab-case or snake_case: `controllers/date_picker.js`, `controllers/list-item.js`).
+In filenames, separate multiple words using either dashes or underscores (kebab-case or snake_case: `controllers/list-item_controller.js`, `controllers/date_picker_controller.js`).
```
javan: Thank you! This is definitely an improvement. Even though mixings of dashes and underscores is supported (`list-item_controller.js`), I think we should be consistent (`list-item-controller.js`). And since we use snake_case more frequently in our docs, I think we should list it first:
```suggestion
In filenames, separate multiple words using either underscores or dashes (snake_case or kebab-case: `controllers/date_picker_controller.js`, `controllers/list-item-controller.js`).
```

javan:
Thanks!

PR #299, opened 2020-03-28
Using Webpacker in Rails

onerinas:
Had to go through different repos and articles to find out there was an easy way to install stimulus for rails. This might help others using Rails project with webpacker.

Issue related to this: https://github.com/stimulusjs/stimulus/issues/59#issuecomment-560507628

### DIFF
--- a/docs/handbook/06_installing_stimulus.md
+++ b/docs/handbook/06_installing_stimulus.md
@@ -58,6 +58,13 @@ const application = Application.start()
 application.register("hello", HelloController)
 application.register("clipboard", ClipboardController)
 ```
+## Using Webpacker in Rails
+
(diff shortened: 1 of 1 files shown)


robzolkos:
Stimulus (and Turbolinks) can (and are) used with non-Rails applications too.    We need to be careful in that we don't unwittingly create a situation where people see only Rails related setup in the README and assume that it is only for Rails.

javan:
Thanks for the suggestion, @onerinas. I agree with @robzolkos's comment above, and this is already [covered by Webpacker's documentation](https://github.com/rails/webpacker/blob/master/docs/integrations.md#stimulus) with a fairly prominent link in its README:

PR #302, opened 2020-04-22
Add a convenientce method for dispatching DOM events inside a controller

adammiribyan:
I came across this method while reading through the javascript code of [Hey](https://hey.com), and have implemented it, along with the globally accessible `ApplicationController`, into a couple of projects already.

I'm only responsible for compiling this PR, credit goes to the folks behind Hey 💯

1. It establishes an adequate event naming convention.
2. It encapsulates the steps you'd have to take to define and emit a custom event into a single `this.dispatch` call.

Here's an simplified e-shop example:
```javascript

// cart_controller.js
  updateCart() {
    // Does stuff...
    this.dispatch('updated', { itemsCount })
  }
}

// cart_badge_controller.js
  static targets = [ 'counter' ]

  initialize() {
    document.addEventListener('cart:updated', this.refreshCounter.bind(this))
  }

  // Actions
  refreshCounter({ detail: { itemsCount } }) {
    this.counterTarget.innerText = itemsCount
  }
}
```

### DIFF
--- a/packages/@stimulus/core/src/context.ts
+++ b/packages/@stimulus/core/src/context.ts
@@ -70,6 +70,13 @@ export class Context implements ErrorHandler {
     return this.element.parentElement
   }
 
+  dispatch(eventName: String, { target = this.element, detail = {}, bubbles = true, cancelable = true } = {}) {
+    const type = `${this.identifier}:${eventName}`
+    const event = new CustomEvent(type, { detail, bubbles, cancelable })
+    target.dispatchEvent(event)
+    return event
+  }
+
   // Error handling
 
   handleError(error: Error, message: string, detail: object = {}) {
--- a/packages/@stimulus/core/src/context.ts
+++ b/packages/@stimulus/core/src/context.ts
@@ -70,8 +70,8 @@ export class Context implements ErrorHandler {
     return this.element.parentElement
   }
 
-  dispatch(eventName: String, { target = this.element, detail = {}, bubbles = true, cancelable = true } = {}) {
-    const type = `${this.identifier}:${eventName}`
+  dispatch(eventName: String, { target = this.element, detail = {}, prefix = this.identifier, bubbles = true, cancelable = true } = {}) {
+    const type = prefix ? `${prefix}:${eventName}` : eventName
     const event = new CustomEvent(type, { detail, bubbles, cancelable })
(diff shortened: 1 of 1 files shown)


timrosskamp:
The `CustomEvent()` constructor is not supported in some older Browsers. There is a small Polyfill [(see mdn)](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill), which should be added to the polyfill package if this gets merged.

dhh:
This is great, @adammiribyan. Could you add some documentation for it as well?

@timrosskamp Support is available in all evergreen browsers, so think we're good there. Although I suppose it'd be nice to document the constraint: https://caniuse.com/?search=CustomEvent

Sub-Xaero:
It would also be nice to see this integrate with the new debug mode added in #354, to give some visibility of events firing whilst debugging.

packages/@stimulus/core/src/context.ts
```diff
@@ -70,6 +70,13 @@ export class Context implements ErrorHandler {
     return this.element.parentElement
   }
 
+  dispatch(eventName: String, { target = this.element, detail = {}, bubbles = true, cancelable = true } = {}) {
+    const type = `${this.identifier}:${eventName}`
```
marcoroth, line 74: I really like this! Maybe we could also make the `prefix` customizable? What do you think?

```suggestion
  dispatch(eventName: String, { target = this.element, detail = {}, prefix = this.identifier, bubbles = true, cancelable = true } = {}) {
    const type = prefix ? `${prefix}:${eventName}` : eventName
```

dancallaghan: Is `prefix` the best name to use here? I've been using `namespace`, but maybe `identifier` is more appropriate?

marcoroth: @dancallaghan I also like `namespace` or `identifier`. 

The reason I suggested  `prefix` is because `stimulus-use` has been using that for a while: https://github.com/stimulus-use/stimulus-use/blob/master/docs/use-dispatch.md#mixin

But I would say that all of the three options work well.

dhh:
Would be nice to get some tests added for this too, if someone wants to start a PR around that.

seb-jean:
I look forward to this feature :)

PR #314, opened 2020-06-30
Specify wrapping data target in controller div

mr-world-wide:
Specify in the handbook that stimulus looks for data-targets defined only inside the controller div, not anywhere in the view.

### DIFF
--- a/docs/handbook/03_building_something_real.md
+++ b/docs/handbook/03_building_something_real.md
@@ -51,7 +51,7 @@ Then add `data-controller="clipboard"` to the outer `<div>`. Any time this attri
 
 ## Defining the Target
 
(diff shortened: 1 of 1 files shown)


docs/handbook/03_building_something_real.md
```diff
@@ -51,7 +51,7 @@ Then add `data-controller="clipboard"` to the outer `<div>`. Any time this attri
 
 ## Defining the Target
 
-We'll need a reference to the text field so we can select its contents before invoking the clipboard API. Add `data-target="clipboard.source"` to the text field:
+We'll need a reference to the text field so we can select its contents before invoking the clipboard API. This data-target must be inside the  <div data-controller defined in the last step.  Add `data-target="clipboard.source"` to the text field:
```
fractaledmind, line 54: This added sentence is a bit malformed. You have an extra space, an opening bracket around the `div` without a closing bracket, and no inline code for the HTML snippet.

PR #318, opened 2020-07-23
Add: Rails example for installing

benkoshy:
## Why this PR?

A subset of folks who use stimulus also use Rails. But how can one easily integrate the two? This minor addition obviates the need for extra searching via stack overflow etc. 

I hope it is helpful.

### DIFF
--- a/docs/handbook/06_installing_stimulus.md
+++ b/docs/handbook/06_installing_stimulus.md
@@ -25,6 +25,17 @@ const application = Application.start()
 const context = require.context("./controllers", true, /\.js$/)
 application.load(definitionsFromContext(context))
 ```
+### Integrating Stimulus with Rails (via Webpacker)
+
+If you are running a Rails 6+ application, getting a working example of Stimulus as simple as hammering: 
+
+`bundle exec rails webpacker:install:stimulus`
+
+This [does the following](https://github.com/rails/webpacker/blob/master/lib/install/stimulus.rb):
+
+* Appends some set up code to your `./app/javascript/packs/application.js` file.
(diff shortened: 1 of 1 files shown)


robzolkos:
Thanks for this @BKSpurgeon  however this has been addressed in https://github.com/stimulusjs/stimulus/pull/299

BKSpurgeon:
The primary concerns, as I see arising from #299 are as follows:

1. People may (incorrectly) assume that Stimulus is only for rails projects, 

**Response:** If that's the case, then I can remove all references to rails and simply use the word: "webpacker". 

2. Webpacker already has the relevant documentation and installation instructions

**Response:** This is a good point: the issue here though is that people might not think to go to the webpacker documentation. 

**New proposal:**

- I'll eliminate all overt references to rails + and explicitly state that rails is not a requirement. your thoughts?

robzolkos:
> * I'll eliminate all overt references to rails

Where are you seeing these References?  I only see two mentions - one in the footer notice ("by Basecamp creators of Ruby on Rails") and one in the history of the project (handbook documentation).    Absent of any overt references to rails being a requirement I don't feel we should explicitly single out Rails as not a requirement?

BKSpurgeon:
@robzolkos apols my language and intent was not clear. I have submitted a revised PR which (if I understand the concerns in #299) will hopefully remedy those concerns: (i) the (incorrect assumption) that one needs Rails in order to use stimulus while balancing the community's demands re: documentation - it seems that it would really benefit users - see below for some examples in the issue you've flagged:

* [Strongly worded comment: ](https://github.com/stimulusjs/stimulus/issues/59#issuecomment-494516683)
* [another comment](https://github.com/stimulusjs/stimulus/issues/59#issue-288384062)

I'll leave it to you to revise/close/merge as you see fit.

rgds

Ben

PR #340, opened 2020-11-25
Account for HTML escaped ActionDescriptor Syntax

seanpdoyle:
Some frameworks (e.g. Rails and [Rack][]) will HTML escape attributes.
This can result in typical action routing attributes to be transformed
and broken:

```html+erb
<button data-action="<%= an_escaping_method("click->controller#action") %>"></button>

<%# => <button data-action="click-&gt;controller#action"></button> %>
```

This commit expands the Action Descriptor syntax to account for
escaping.

In the case of Rack, there is a small collection of HTML escaped
characters:

* `&` escapes into `&amp;`
* `<` escapes into `&lt;`
* `>` escapes into `&gt;`
* `'` escapes into `&#x27;`
* `"` escapes into `&quot;`
* `/` escapes into `&#x2F;`

Of those characters, the `>` is the only one currently part of the
Stimulus syntax.

[Rack]: https://www.rubydoc.info/github/rack/rack/Rack/Utils#ESCAPE_HTML-constant

### DIFF
--- a/packages/@stimulus/core/src/action_descriptor.ts
+++ b/packages/@stimulus/core/src/action_descriptor.ts
@@ -6,8 +6,8 @@ export interface ActionDescriptor {
   methodName: string
 }
 
-// capture nos.:            12   23 4               43 5        51 6   67 8      879 X  X9
-const descriptorPattern = /^((.+?)(@(window|document))?(->|-&gt;))?(.+?)(#([^:]+?))(:(.+))?$/
+// capture nos.:            12   23 4               43   1 5   56 7      768 9  98
+const descriptorPattern = /^((.+?)(@(window|document))?->)?(.+?)(#([^:]+?))(:(.+))?$/
 
 export function parseDescriptorString(descriptorString: string): Partial<ActionDescriptor> {
   const source = descriptorString.trim()
@@ -15,9 +15,9 @@ export function parseDescriptorString(descriptorString: string): Partial<ActionD
   return {
     eventTarget:  parseEventTarget(matches[4]),
     eventName:    matches[2],
-    eventOptions: matches[10] ? parseEventOptions(matches[10]) : {},
-    identifier:   matches[6],
-    methodName:   matches[8]
+    eventOptions: matches[9] ? parseEventOptions(matches[9]) : {},
+    identifier:   matches[5],
+    methodName:   matches[7]
   }
 }
 
--- a/packages/@stimulus/core/src/action_descriptor.ts
+++ b/packages/@stimulus/core/src/action_descriptor.ts
@@ -6,8 +6,8 @@ export interface ActionDescriptor {
   methodName: string
 }
 
-// capture nos.:            12   23 4               43   1 5   56 7      768 9  98
-const descriptorPattern = /^((.+?)(@(window|document))?->)?(.+?)(#([^:]+?))(:(.+))?$/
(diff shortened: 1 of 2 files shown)


packages/@stimulus/core/src/action_descriptor.ts
```diff
@@ -7,7 +7,7 @@ export interface ActionDescriptor {
 }
 
 // capture nos.:            12   23 4               43   1 5   56 7      768 9  98
-const descriptorPattern = /^((.+?)(@(window|document))?->)?(.+?)(#([^:]+?))(:(.+))?$/
+const descriptorPattern = /^((.+?)(@(window|document))?(->|&gt;))?(.+?)(#([^:]+?))(:(.+))?$/
```
seanpdoyle, line 10: I couldn't regexp myself out of a wet paper bag, and I wonder if this style disrupts the `// capture nos.:` above.

seanpdoyle:
As an alternative to this, we could change the frameworks to be less strict and skip the escaping, but slightly expanding the Stimulus syntax feels like a lower risk solution.

What do you think @sstephenson @javan?

packages/@stimulus/core/src/action_descriptor.ts
```diff
@@ -7,7 +7,7 @@ export interface ActionDescriptor {
 }
 
 // capture nos.:            12   23 4               43   1 5   56 7      768 9  98
-const descriptorPattern = /^((.+?)(@(window|document))?->)?(.+?)(#([^:]+?))(:(.+))?$/
+const descriptorPattern = /^((.+?)(@(window|document))?(->|-&gt;))?(.+?)(#([^:]+?))(:(.+))?$/
```
seanpdoyle, line 10: I couldn't regexp myself out of a wet paper bag, and I wonder if this style disrupts the `// capture nos.:` above.

sstephenson:
Could you clarify how the action attributes are broken? Or is this just an aesthetic issue? This example works fine:

```html
<!DOCTYPE html>
<html>
  <head>
    <script src="https://unpkg.com/stimulus"></script>
    <script>
      const app = new Stimulus.Application
      app.register("hello", class extends Stimulus.Controller {
        greet() {
          alert("Hello!")
        }
      })
      app.start()
    </script>
  </head>
  <body>
    <button data-controller="hello" data-action="click-&gt;hello#greet">Greet</button>
  </body>
</html>
```

seanpdoyle:
@sstephenson wow, thanks for pointing that out. I've pushed up a commit to run the test without the implementation changes.

Re-visiting the source error that motivated this change, it appears that I'm doubly escaping the HTML: `->` becomes `&gt;`, and then `&gt;` becomes `&amp;gt;`. If that's the true root of this issue, I don't think it's Stimulus' responsibility to account for that.

seanpdoyle:
The suite passes without the implementation change.

@sstephenson if you don't think this test is a valuable addition to the suite, let's close this PR.

sstephenson:
Don't think we need an explicit test for this, since it's testing the web platform itself, not Stimulus. 👍

I do think it's interesting to consider how HTML escaping further up the stack interacts with our action descriptor syntax. Technically there's no reason to escape \> in a quoted attribute value, but most implementations do it anyway to reduce the attack surface for XSS, or because they don't have context about where the escaped string is used (like in your ERB example). It would be nice if we had better tooling for emitting unescaped \>s, but either way it's just an aesthetic thing and ultimately has no bearing on how Stimulus sees the action descriptors.

PR #343, opened 2020-12-07
Update the target not found error message to reflect the new Target syntax

adrienpoly:
Small detail but the error message when a target is not found uses the old target syntax
```js
new Error(`Missing target element "${this.identifier}.${name}"`)
```
Screen Shot 2020-12-07 at 9 56 29 PM

This PR changes the message to 

```js
new Error(`Missing target element : data-${this.identifier}-target="${name}"`)
```

Screen Shot 2020-12-07 at 9 58 56 PM

### DIFF
--- a/packages/@stimulus/core/src/target_properties.ts
+++ b/packages/@stimulus/core/src/target_properties.ts
@@ -19,7 +19,7 @@ function propertiesForTargetDefinition(name: string) {
         if (target) {
           return target
         } else {
-          throw new Error(`Missing target element "${this.identifier}.${name}"`)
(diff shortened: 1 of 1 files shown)


javan:
Thanks for catching this, @adrienpoly! What do you think about removing the target syntax entirely to make legacy / future compatibility less of an issue? Something like:
```js
new Error(`Missing target element "${name}" for "${this.identifier}" controller`)
```

adrienpoly:
Excellent suggestion. The message is updated accordingly.

Out of the scope of this PR but maybe errors could have a link to the reference  page of the handbook. 

Also, not always easy to maintain unless everything is centralized in one place. 

Maybe the `guide.ts` could centralize those messages.

PR #344, opened 2020-12-08
Add [key]Classes method to better handle multiple CSS classes

### LINKED ISSUE #341: Add [logicalName]Classes
(extracted from https://github.com/stimulusjs/stimulus/pull/202#issuecomment-620553981)

To ease working with CSS Utility frameworks (such as [Tailwind](https://tailwindcss.com/)), I'm proposing adding `this.[logicalName]Classes` that would be equivalent to `this.[name]Class.split(' ')`. You can then use it with the spread operator to add and remove classes.

```html
<div data-controller="clipboard"
     data-clipboard-success-class="bg-green-500 font-bold">
```
```js
// clipboard_controller.js

  initialize() {
    if (/* ... */) {
      this.element.classList.add(this.supportedClass)
    }
  }

  copy() {
    // ...
    this.element.classList.add(...this.successClasses)
  }
}
```

I'm willing to take a stab at implementing it if you're interested in the concept.

And thanks for this wonderful library that shows there are more than one way to build modern websites.

swanson:
Closes #341

:wave: 

Utility-first CSS is becoming [increasingly popular](https://2020.stateofcss.com/en-US/technologies/methodologies/#css_methodologies_experience_ranking) as an alternative to BEM (the convention used in the Handbook). This PR adds the `[key]Classes` property that maps a space separated list of classes into an array.

This new property can be combined with the spread operator for more elegant `classList` manipulation when you need to change multiple classes at once.

Here is an adapted example from my production codebase:

```js
// highlight_controller.js
  static targets = [ "subject" ]
  static classes = [ "pulse" ]

  pulse() {
    this.subjectTarget.classList.add(...this.pulseClasses)
  }
}
```

```html
<div data-controller="highlight" 
     data-highlight-pulse-class="animate-pulse ring-8 ring-teal-600 ring-offset-8 rounded">

  <button data-action="click->highlight#pulse">
    Show me
  </button>

  <span data-highlight-target="subject">
     ...
  </span>
</div>
```

Pros:
- Cleaner usage with utility CSS (e.g. Tailwind)
- `${key}Classes` plural API mirrors target conventions

Cons:
- Inconsistency that `fooTargets` is mapped by multiple `data-x-y-targets` attributes, while `fooClasses` is mapped by space separated string (too much magic?)

I've added some basic test cases and happy to add docs if this is something you'd like added. Open to any and all feedback or alternatives :)

### DIFF
--- a/docs/reference/css_classes.md
+++ b/docs/reference/css_classes.md
@@ -64,7 +64,7 @@ Construct a CSS class attribute by joining together the controller identifier an
 </form>
 ```
 
-If you are working with utility classes and want to define multiple CSS classes for an attribute, separate the CSS class names with spaces.
+If you want to define multiple CSS classes for an attribute, separate the CSS class names with spaces.
 
 ## Properties
 
@@ -96,7 +96,7 @@ export default class extends Controller {
 
 **Note:** Stimulus will throw an error if you attempt to access a CSS class property when no matching CSS class attribute is present.
 
-If you are using utility classes, access the plural class property. You can apply the array of CSS classes to elements using [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).
+If you want to use multiple classes, access the plural class property. You can apply the array of CSS classes to elements using [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).
 
 <meta data-controller="callout" data-callout-text-value="...this.loadingClasses">
 
--- a/docs/reference/css_classes.md
+++ b/docs/reference/css_classes.md
@@ -112,7 +112,6 @@ export default class extends Controller {
 }
 ```
 
-
 ## Naming Conventions
 
 Use camelCase to specify logical names in CSS class definitions. Logical names map to camelCase CSS class properties:
--- a/docs/reference/css_classes.md
+++ b/docs/reference/css_classes.md
@@ -55,14 +55,26 @@ Construct a CSS class attribute by joining together the controller identifier an
 
 **Note:** CSS class attributes must be specified on the same element as the `data-controller` attribute.
 
+<meta data-controller="callout" data-callout-text-value="data-search-loading-class=&quot;bg-gray-500 animate-spinner cursor-busy&quot;">
+
+```html
+<form data-controller="search"
+      data-search-loading-class="bg-gray-500 animate-spinner cursor-busy">
+  <input data-action="search#loadResults">
+</form>
+```
+
+If you are working with utility classes and want to define multiple CSS classes for an attribute, separate the CSS class names with spaces.
+
 ## Properties
 
 For each logical name defined in the `static classes` array, Stimulus adds the following _CSS class properties_ to your controller:
 
-Name                    | Value
------------------------ | -----
-`[logicalName]Class`    | The value of the CSS class attribute corresponding to `logicalName`
-`has[LogicalName]Class` | A boolean indicating whether or not the CSS class attribute is present
+Kind        | Name                         | Value
+----------- | ---------------------------- | -----
+Singular    | `this.[logicalName]Class`    | The value of the CSS class attribute corresponding to `logicalName`
+Plural      | `this.[logicalName]Classes`  | An array of all CSS class attributes, split by spaces
+Existential | `this.has[LogicalName]Class` | A boolean indicating whether or not the CSS class attribute is present
 
 <br>Use these properties to apply CSS classes to elements with the `add()` and `remove()` methods of the [DOM `classList` API](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList).
 
@@ -83,6 +95,23 @@ export default class extends Controller {
 ```
 
 **Note:** Stimulus will throw an error if you attempt to access a CSS class property when no matching CSS class attribute is present.
+
+If you are using utility classes, access the plural class property. You can apply the array of CSS classes to elements using [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).
+
+<meta data-controller="callout" data-callout-text-value="...this.loadingClasses">
+
+```js
+export default class extends Controller {
+  static classes = [ "loading" ]
+
+  loadResults() {
+    this.element.classList.add(...this.loadingClasses)
+
+    fetch(/* … */)
+  }
+}
+```
+
 
 ## Naming Conventions
 
--- a/packages/@stimulus/core/src/class_map.ts
+++ b/packages/@stimulus/core/src/class_map.ts
@@ -1,4 +1,5 @@
 import { Scope } from "./scope"
+import { tokenize } from "./string_helpers"
 
 export class ClassMap {
   readonly scope: Scope
@@ -17,7 +18,7 @@ export class ClassMap {
 
   getAll(name: string) {
     const tokenString = this.data.get(this.getDataKey(name)) || ""
-    return tokenString.trim().split(/\s+/).filter(content => content.length)
+    return tokenize(tokenString)
   }
 
   getAttributeName(name: string) {
--- a/packages/@stimulus/core/src/class_map.ts
+++ b/packages/@stimulus/core/src/class_map.ts
@@ -12,7 +12,11 @@ export class ClassMap {
   }
 
   get(name: string) {
-    return this.data.get(this.getDataKey(name))
+    return this.getAll(name)[0];
+  }
+
+  getAll(name: string) {
+    return this.data.get(this.getDataKey(name))?.split(" ") || []
   }
 
   getAttributeName(name: string) {
--- a/packages/@stimulus/core/src/class_map.ts
+++ b/packages/@stimulus/core/src/class_map.ts
@@ -12,17 +12,12 @@ export class ClassMap {
   }
 
   get(name: string) {
-    return this.getAll(name)[0];
+    return this.getAll(name)[0]
   }
 
   getAll(name: string) {
-    const classAttribute = this.data.get(this.getDataKey(name));
-
-    if (classAttribute) {
-      return classAttribute.split(" ")
-    } else {
-      return []
-    }
+    const tokenString = this.data.get(this.getDataKey(name)) || ""
+    return tokenString.trim().split(/\s+/).filter(content => content.length)
   }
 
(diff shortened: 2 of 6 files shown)


packages/@stimulus/core/src/class_properties.ts
```diff
@@ -27,7 +25,13 @@ function propertiesForClassDefinition(key: string) {
       }
     },
 
-    [`has${capitalize(name)}`]: {
+    [`${key}Classes`]: {
+      get(this: Controller) {
+        return this.classes.get(key)?.split(" ") || []
```
swanson, line 30: I'm new to typescript so if there is a better way to "annotate" this, please let me know :)

swanson:
Would appreciate any input on this proposal @dhh -- thanks!

javan:
Makes sense to support utility-first CSS conventions, but overloading the same data attribute to mean one or many class names seems problematic. If you can't trust `this.fooClass` to return a single class name then you can no longer do `element.classList.add(this.fooClass)` safely. Updating the singular property to return the plural property's first item might work. That's more or less what target properties do under the hood.

dhh:
@javan You don't like having fooClass just return the string and then fooClasses return an array where split has been called on the string? Then they're just two types of accessors. One always returns a string, the other always returns an array (even if it might only have one element, if there are no spaces).

swanson:
> Makes sense to support utility-first CSS conventions, but overloading the same data attribute to mean one or many class names seems problematic. If you can't trust `this.fooClass` to return a single class name then you can no longer do `element.classList.add(this.fooClass)` safely. Updating the singular property to return the plural property's first item might work. That's more or less what target properties do under the hood.

Yeah, I wasn't sure if it was preferable to have the singular property blow up if you pass a string with spaces (to notify you that you very likely have a typo/bug) or have it silently not apply anything other than the first element.

Is throwing a Stimulus error (not the raw `DomException`) an option? Personally, I think I'd rather have that happen but I can see not wanting to minimize possible errors in the library.

javan:
> You don't like having fooClass just return the string and then fooClasses return an array where split has been called on the string?

Using the example above:
```html
<div data-controller="highlight" 
     data-highlight-pulse-class="animate-pulse ring-8 ring-teal-600 ring-offset-8 rounded">
```
I don't like `this.pulseClass` returning the full string `"animate-pulse ring-8 ring-teal-600 ring-offset-8 rounded"` because it's not a singular class name and will throw an exception if you `classList.add(this.pulseClass)`. I think it should return the *first* class instead, `"animate-pulse"`. Just like targets.

dhh:
Yeah, that's a great point. I buy that.

javan:
> Is throwing a Stimulus error (not the raw DomException) an option? Personally, I think I'd rather have that happen but I can see wanting to minimize possible errors in the library.

Singular properties should throw an error. They do currently: https://github.com/stimulusjs/stimulus/blob/11370bc37ac07753dbc4d676eebb8534bb03d3f8/packages/%40stimulus/core/src/class_properties.ts#L18-L26

Unsure if plural properties should do the same.

swanson:
> 
> Singular properties should throw an error. They do currently:
> 

Sorry, I was unclear. I meant should we throw a different error ("Expected a string without spaces, did you mean `fooClasses`?"). Sounds like returning the first class was the preferred option.

> 
> Unsure if plural properties should do the same.
>

Yeah, I'm not sure either. Plural targets do not check for missing the key (maybe that is intentional to support 0..N targets, not 1..N targets?).

But in the utility CSS use-case, it would be nice if Stimulus errored if you typo the key or omit it. You're unlikely to use the singular property if you are expecting multiple classes so you are skipping the missing attribute checks.

Trying to think if there is any use-case where you would intentionally want to pass in an empty string for the class value...

docs/reference/css_classes.md
```diff
@@ -55,14 +55,26 @@ Construct a CSS class attribute by joining together the controller identifier an
...
+<meta data-controller="callout" data-callout-text-value="data-search-loading-class=&quot;bg-gray-500 animate-spinner cursor-busy&quot;">
+
+```html
+<form data-controller="search"
+      data-search-loading-class="bg-gray-500 animate-spinner cursor-busy">
+  <input data-action="search#loadResults">
+</form>
+```
+
+If you are working with utility classes and want to define multiple CSS classes for an attribute, separate the CSS class names with spaces.
```
javan, line 67: Let's drop "utility classes" from the docs since it's not the *only* reason for this change and is more of a convention. Fine to just frame it around "multiple classes".

packages/@stimulus/core/src/tests/modules/class_tests.ts
```diff
@@ -6,21 +6,32 @@ export default class ValueTests extends ControllerTestCase(ClassController) {
...
     this.assert.equal(this.controller.loadingClass, "busy")
   }
+
+  "test utility classes map to array"() {
+    this.assert.deepEqual(this.controller.successClasses, ["bg-green-400", "border", "border-green-600"])
+  }
+
+  "test accessing a class property returns first class if utility classes are used"() {
+    this.assert.equal(this.controller.successClass, "bg-green-400");
+  }
```
javan, line 36: Ditto, replace "utility" with "multiple"

packages/@stimulus/core/src/class_map.ts
```diff
@@ -12,7 +12,17 @@ export class ClassMap {
   }
 
   get(name: string) {
-    return this.data.get(this.getDataKey(name))
+    return this.getAll(name)[0];
```
javan, line 15: ✂️ `;` to match existing coding style (we don't use them)

packages/@stimulus/core/src/class_map.ts
```diff
@@ -12,7 +12,17 @@ export class ClassMap {
...
+  }
+
+  getAll(name: string) {
+    const classAttribute = this.data.get(this.getDataKey(name));
```
javan, line 19: ✂️ `;`

packages/@stimulus/core/src/class_map.ts
```diff
@@ -12,7 +12,17 @@ export class ClassMap {
...
+
+    if (classAttribute) {
+      return classAttribute.split(" ")
```
javan, line 22: `tokenString` might be a better name for this variable, and I'd trim + split on white space. Following the implementation here https://github.com/stimulusjs/stimulus/blob/8cb58d1f8a4ac83a875c7a1fdc7a28f21869efd6/packages/%40stimulus/mutation-observers/src/token_list_observer.ts#L106 this method could be rewritten as
```js
getAll(name: string) {
  const tokenString = this.data.get(this.getDataKey(name)) || ""
  return tokenString.trim().split(/\s+/)
}
```

javan (changes requested):
Thanks for taking this on! Requesting a few minor changes.

packages/@stimulus/core/src/class_map.ts
swanson: > this method could be rewritten as
> ```js
> getAll(name: string) {
>   const tokenString = this.data.get(this.getDataKey(name)) || ""
>   return tokenString.trim().split(/\s+/)
> }
> ```

Good feedback -- just to clarify, we'll still need the `filter` otherwise we will return `[""]` if the controller is missing the class attribute.

packages/@stimulus/core/src/class_map.ts
```diff
@@ -12,7 +12,12 @@ export class ClassMap {
   }
 
   get(name: string) {
-    return this.data.get(this.getDataKey(name))
+    return this.getAll(name)[0]
+  }
+
+  getAll(name: string) {
+    const tokenString = this.data.get(this.getDataKey(name)) || ""
+    return tokenString.trim().split(/\s+/).filter(content => content.length)
```
swanson, line 20: Should this be refactored out to `string_helpers` as something like `splitTokens`? Or wait until three usages?

swanson:
@javan thank you for reviewing. I've made the requested changes.

packages/@stimulus/core/src/class_map.ts
swanson: I went ahead and pulled this out into a function `tokenize` in `string_helpers.ts`. I did not update similar code in the `token_list_observer` as it lives in a separate package and is unrelated to the new functionality in this PR.

swanson:
Holler if there is anything else we want to do with this, I have some spare time during the holidays. Otherwise, I think it's ready for final review/merge. I can imagine the past few weeks have been pretty hectic so I appreciate the time to review this!

javan:
Thanks, @swanson! We got a little sidelined getting hotwire out the door, and most of us are out for holidays now. Will resume soon!

swanson:
:wave: friendly bump :)

sstephenson:
This is at the top of our list when we return to Stimulus later this month 👍

sstephenson:
Wonderful work on this all around, @swanson! Merged with a few revisions to the documentation: ce942f9

janko:
For anyone wanting to use this functionality while waiting on the next Stimulus release, this is what I came up with:

```rb
// app/javascripts/controllers/application_controller.js

  // TODO: Remove when https://github.com/hotwired/stimulus/pull/344 gets released.
  initialize() {
    this.constructor.classes.forEach(logicalClass => {
      Object.defineProperty(this, `${logicalClass}Classes`, {
        get() {
          return this.[`${logicalClass}Class`].split(' ')
        }
      })
    })
  }
}
```

Other controllers can subclass from this controller.

marcoroth:
You can also swap the `stimulus` dependency in your `package.json` to the latest dev-build, like so:
```js
"stimulus": "https://github.com/hotwired/dev-builds/archive/stimulus/848e0c7.tar.gz"
```

packages/@stimulus/core/src/string_helpers.ts
```diff
@@ -9,3 +9,7 @@ export function capitalize(value: string) {
 export function dasherize(value: string) {
   return value.replace(/([A-Z])/g, (_, char) => `-${char.toLowerCase()}`)
 }
+
+export function tokenize(value: string) {
+  return value.trim().split(/\s+/).filter(content => content.length)
```
tleish, line 14: A simpler (and slightly faster) method of doing this is
```javascript
return value.match(/[^ ]+/g)
```

dhh: Please do add a PR 👍

tleish: PR https://github.com/hotwired/stimulus/pull/430

PR #351, opened 2020-12-15
Add CSS escaping for query selectors

tgalopin:
Hi! First PR on Stimulus, hopefully not the last :) !

In the context of [Symfony UX](https://symfony.com/blog/new-in-symfony-the-ux-initiative-a-new-javascript-ecosystem-for-symfony), we rely on the `@` character to note a controller coming from the vendors (for instance:  `@symfony/ux-chartjs/chart`).

In Stimulus 1, this notation was fine because targets were defined using the legacy notation. With the new notation, the `@` and `/` raise issues (which makes sense).

I'm working on migrating away from `/` to `--`, as is natively done by Stimulus Webpack helpers. However, `@` is fine in HTML attribute names, and I'd like to keep it as it would allow us to detect controllers coming from vendors vs the ones coming from the app.

For the moment, the main problem I see is that Stimulus doesn't escape CSS selectors when using them, thus this PR. CSS.escape is supported by many majors browsers (https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) and a reliable polyfill exists (https://github.com/mathiasbynens/CSS.escape), which we could add as a peerDependency if you find it useful.

I'm not sure how to test this for now, if you have any tip I'd be happy to learn!

### DIFF
--- a/packages/@stimulus/core/src/scope.ts
+++ b/packages/@stimulus/core/src/scope.ts
@@ -40,7 +40,7 @@ export class Scope {
   }
 
   private queryElements(selector: string): Element[] {
-    return Array.from(this.element.querySelectorAll(selector))
+    return Array.from(this.element.querySelectorAll(CSS.escape(selector)))
   }
 
   private get controllerSelector(): string {
--- a/packages/@stimulus/polyfills/index.js
+++ b/packages/@stimulus/polyfills/index.js
@@ -9,6 +9,7 @@ import "core-js/es/string/starts-with"
 import "element-closest"
 import "mutation-observer-inner-html-shim"
 import "eventlistener-polyfill"
+import "css.escape"
 
 if (typeof SVGElement.prototype.contains != "function") {
   SVGElement.prototype.contains = function(node) {
--- a/packages/@stimulus/polyfills/package.json
+++ b/packages/@stimulus/polyfills/package.json
@@ -11,6 +11,7 @@
   "license": "MIT",
   "dependencies": {
     "core-js": "^3.4.0",
+    "css.escape": "^1.5.1",
     "element-closest": "^2.0.2",
     "eventlistener-polyfill": "^1.0.5",
     "mutation-observer-inner-html-shim": "^1.0.0"


javan:
Thanks! +1 to this change, generally.

I think we should add the polyfill as a dependency of [`@stimulus/polyfills`](https://github.com/stimulusjs/stimulus/tree/master/packages/%40stimulus/polyfills). Doing so should fix the failing tests on CI.

tgalopin:
Done!

packages/@stimulus/core/src/scope.ts
```diff
@@ -40,7 +40,7 @@ export class Scope {
   }
 
   private queryElements(selector: string): Element[] {
-    return Array.from(this.element.querySelectorAll(selector))
+    return Array.from(this.element.querySelectorAll(CSS.escape(selector)))
```
javan, line 43: I don't think escaping the entire selector fixes the problem. In fact, it mostly breaks things. Here's a simplified example:
```js
> selector = `[data-controller~="hello"]`
"[data-controller~=\"hello\"]"

> escapedSelector = CSS.escape(selector)
"\\[data-controller\\~\\=\\\"hello\\\"\\]"

> document.querySelector(selector)
<div class="landing-hello" data-controller="hello">

> document.querySelector(escapedSelector)
null
```

`CSS.escape()` will need to be targeted more tightly within the selector.

javan (changes requested):

packages/@stimulus/core/src/scope.ts
tgalopin: I could try to add CSS.espace only on the attribute names, where they are defined (targets, actions, controllers, ...). WDYT?

javan: I'd like the test suite help guide you. `yarn test` locally or, rebase this branch to pick up latest CI changes so you can see the test results.

tgalopin:
I experimented a little more and it seems browsers don't like `@` in attributes names, even if they aren't "officially" disallowed. To avoid issues, I'll find a better way to fix this in Symfony UX.

Thanks for the help anyway! I'm glad to have discovered the Stimulus codebase :)

Issue #349, opened 2020-12-11
[Enhancement proposal] debug mode with lifecycle events and action logs


adrienpoly:
I find my self very often adding console.log when building Stimulus controllers. I use them for many things but mostly
- make sure my controller is properly loaded (no spelling mistakes) 
- log actions to confirm the event is correctly setup and/or inspect the payload
- check my target lists etc

Talking with others (ping @ssaunier, @db0sch) we thought in terms of DX it would be nice to have a more generic way of doing this and potentially hook it to log service such as logRocket for debugging

That it something I started to implement for the [StimulusUse](https://github.com/stimulus-use/stimulus-use) mixins ([example with useDispatch](https://github.com/stimulus-use/stimulus-use/pull/85)).

I looked whether this could be some kind of `useDebug` mixin in StimulusUse to optionally enhance Stimulus without bloating the core with it. I think it would be possible **but with quite a bit of monkey patching and using non official API** before looking into doing something outside I wanted to submit a proposal of API as a core Stimulus feature to discuss.

## Draft API

Here is a proposal of what it could look like

### Setting

Either 
```js
application.debug(process.env.NODE_ENV === 'development')
```

or 

```js
application.start(debug: process.env.NODE_ENV === 'development') // I think this would be a breaking change for the start function
```

### Logs

The following would be logged:
- initialize/connect/disconnect lifecycle
- in the `initialize` collapse add the target elements (or simply count) 
- actions
- [bonus] valueChanged

use a [groupCollapsed](https://developer.mozilla.org/fr/docs/Web/API/Console/groupCollapsed) with the lifecycle/action name and some context details in the collapsed zone.

Example:

![Screen Shot 2020-12-11 at 3 59 09 PM](https://user-images.githubusercontent.com/7847244/101918608-d8669980-3bc9-11eb-9b4f-7831c04db300.jpg)

This should leverage the current Logger class to combine to a single point all error/warning and log activities.

Would this be something that you would be interested to have in terms of core feature ?

At the core feature level it could maybe be as simple as having a hook at the application level for all activities and this hook can then be customized.

dhh:
Would be great to get debugging, yes. How about it just being a flag that can be toggled:

```javascript
if (process.env.NODE_ENV === 'development') {
  application.debug = true
}
```

Then you could even toggle it at runtime if your application object is accessible from the console.

adrienpoly:
Yes this was exactly my idea with a slightly different syntax (function rather than setter) but both work.

[...]

PR #354, opened 2020-12-17
Introduce a Debug mode

adrienpoly:
This PR adds a debug mode to Stimulus and close [#349](https://github.com/stimulusjs/stimulus/issues/349)

In debug mode _application lifecycles_, _controller lifecycles_ and _actions_ events are sent to the Stimulus `Logger` with a default output to the console. Each log includes context information available in a detailed view.

**Example:**
Kapture 2020-12-17 at 11 37 18

## Usage

debug mode can be enabled during the Stimulus application initialization phase

```js
import { Application } from 'stimulus'
import { definitionsFromContext } from 'stimulus/webpack-helpers'

const context = require.context('./controllers', true, /\.js$/)

// enable Stimulus debug mode in development
application.debug = process.env.NODE_ENV === 'development'
```

Or even at run time as suggested [here](https://github.com/stimulusjs/stimulus/issues/349#issuecomment-743285759) if `application` is accessible at the window level.

```js
window.stimulusApplication = application

// and then in the console 
stimulusApplication.debug = true // or false to toggle the debug mode in production
```

#### Content

Messages are collapsed groups. Title of the group follow the pattern `emitter #action`
The emitter can be `application` or a `controller identifier`

**Examples:**
- `application #staring`
- `application #stared`
- `form #initialize`
- `form #connect`
- `form #submit` -> action

Opening the group provides a context object with details:

- application (all)
- controller (controller lifecycle and actions)
- element (controller lifecycle and actions)
- identifier (controller lifecycle and actions)
- functionName (actions)
- currentTarget (actions)
- event (actions)
- target (actions)

#### Style

I applied a few colors with a variant for dark mode. I took Purple and Yellow from Basecamp website. Matter of taste and easily changeable.

|light mode|dark mode|
|---|---|
|Screen Shot 2020-12-17 at 11 13 42 AM|Screen Shot 2020-12-17 at 11 12 54 AM|

## Values

Value changes are not logged at this point. It shouldn't be difficult to add if needed.

## Namespacing

No particular namespace is applied to the log message at this point. This is a subject of discussion. Do we want something like : 

`Stimulus - emitter #action` for the collapsed group title ?

### Custom logger

By default, the logger is the `console`. A custom `Logger`can be supply

```js
application.logger = customLogger
```

### Tests

This PR does not include tests at this point. I am not sure how you would like this to be tested. One approach I could think of is to create a custom Logger that would collect messages (could store them in the `application`). And then do some assertion on the amount of messages and type we receive. An approach a bit similar to the `log_controller_test_case` but at the application level this time.

### DIFF
--- a/packages/@stimulus/core/src/application.ts
+++ b/packages/@stimulus/core/src/application.ts
@@ -32,14 +32,14 @@ export class Application implements ErrorHandler {
     this.logDebugActivity("application", "starting")
     this.dispatcher.start()
     this.router.start()
-    this.logDebugActivity("application", "started")
+    this.logDebugActivity("application", "start")
   }
 
   stop() {
     this.logDebugActivity("application", "stopping")
     this.dispatcher.stop()
     this.router.stop()
-    this.logDebugActivity("application", "stoped")
+    this.logDebugActivity("application", "stop")
   }
 
   register(identifier: string, controllerConstructor: ControllerConstructor) {
--- a/packages/@stimulus/core/src/application.ts
+++ b/packages/@stimulus/core/src/application.ts
@@ -36,7 +36,7 @@ export class Application implements ErrorHandler {
   }
 
   stop() {
-    this.logDebugActivity("application", "stoping")
+    this.logDebugActivity("application", "stopping")
     this.dispatcher.stop()
     this.router.stop()
     this.logDebugActivity("application", "stoped")
--- a/packages/@stimulus/core/src/application.ts
+++ b/packages/@stimulus/core/src/application.ts
@@ -12,6 +12,7 @@ export class Application implements ErrorHandler {
   readonly dispatcher: Dispatcher
   readonly router: Router
   logger: Logger = console
+  debug: boolean = false
 
   static start(element?: Element, schema?: Schema): Application {
     const application = new Application(element, schema)
@@ -28,13 +29,17 @@ export class Application implements ErrorHandler {
 
   async start() {
     await domReady()
+    this.logDebugActivity("application", "starting")
     this.dispatcher.start()
     this.router.start()
+    this.logDebugActivity("application", "started")
   }
 
   stop() {
+    this.logDebugActivity("application", "stoping")
     this.dispatcher.stop()
     this.router.stop()
+    this.logDebugActivity("application", "stoped")
   }
 
   register(identifier: string, controllerConstructor: ControllerConstructor) {
@@ -70,6 +75,24 @@ export class Application implements ErrorHandler {
 
   handleError(error: Error, message: string, detail: object) {
     this.logger.error(`%s\n\n%o\n\n%o`, message, error, detail)
+  }
+
+  // Debug logging
+
+  logDebugActivity = (identifier: string, functionName: string, detail: object = {}): void => {
+    if (this.debug) {
+      this.logFormattedMessage(identifier, functionName, detail)
+    }
+  }
+
+  private logFormattedMessage(identifier: string, functionName: string, detail: object = {}) {
+    const darkMode = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+    const color = darkMode ? "#ffe000" : "#5D2F85"
+    detail = Object.assign({ application: this }, detail)
+
+    this.logger.groupCollapsed(`%c${identifier}%c #${functionName}`, `color: ${color}`, 'color: unset')
+    this.logger.log("details:", { ...detail })
+    this.logger.groupEnd()
   }
 }
 
--- a/packages/@stimulus/core/src/binding.ts
+++ b/packages/@stimulus/core/src/binding.ts
@@ -47,8 +47,10 @@ export class Binding {
   }
 
   private invokeWithEvent(event: Event) {
+    const { target, currentTarget } = event
     try {
       this.method.call(this.controller, event)
+      this.context.logDebugActivity(this.methodName, { event, target, currentTarget, action: this.methodName })
     } catch (error) {
       const { identifier, controller, element, index } = this
       const detail = { identifier, controller, element, index, event }
--- a/packages/@stimulus/core/src/context.ts
+++ b/packages/@stimulus/core/src/context.ts
@@ -24,6 +24,7 @@ export class Context implements ErrorHandler {
 
     try {
       this.controller.initialize()
+      this.logDebugActivity("initialize")
     } catch (error) {
       this.handleError(error, "initializing controller")
     }
@@ -35,6 +36,7 @@ export class Context implements ErrorHandler {
 
     try {
       this.controller.connect()
+      this.logDebugActivity("connect")
     } catch (error) {
       this.handleError(error, "connecting controller")
     }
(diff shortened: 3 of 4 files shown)


packages/@stimulus/core/src/application.ts
```diff
@@ -28,13 +29,17 @@ export class Application implements ErrorHandler {
...
   async start() {
     await domReady()
+    this.logDebugActivity("application", "starting")
     this.dispatcher.start()
     this.router.start()
+    this.logDebugActivity("application", "started")
   }
 
   stop() {
+    this.logDebugActivity("application", "stoping")
```
db0sch, line 39:
```suggestion
    this.logDebugActivity("application", "stopping")
```

source: https://wikidiff.com/stoped/stopped

packages/@stimulus/core/src/application.ts
```diff
@@ -28,13 +29,17 @@ export class Application implements ErrorHandler {
...
     this.dispatcher.stop()
     this.router.stop()
+    this.logDebugActivity("application", "stoped")
```
db0sch, line 42:
```suggestion
    this.logDebugActivity("application", "stopped")
```

packages/@stimulus/core/src/application.ts
```diff
@@ -71,6 +76,24 @@ export class Application implements ErrorHandler {
...
+  // Debug logging
+
+  logDebugActivity = (identifier: string, functionName: string, detail: object = {}): void => {
+    if (this.debug) {
+      this.logFormattedMessage(identifier, functionName, detail)
+    }
+  }
+
+  private logFormattedMessage(identifier: string, functionName: string, detail: object = {}) {
+    const darkMode = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
```
db0sch, line 89: can we use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here to shorten the condition?

packages/@stimulus/core/src/application.ts
```diff
@@ -28,13 +29,17 @@ export class Application implements ErrorHandler {
 
   async start() {
     await domReady()
+    this.logDebugActivity("application", "starting")
     this.dispatcher.start()
     this.router.start()
+    this.logDebugActivity("application", "started")
```
db0sch, line 35: All other debug logs use the infinite tense (`initialize`, `connect`), not here.
Maybe we can harmonize and just do `this.logDebugActivity("application", "start")`

db0sch (commented):
Thanks @adrienpoly for implementing this.
No more `console.log`  in `connect()` 😉 

Just a couple of small comments.

packages/@stimulus/core/src/application.ts
adrienpoly: I can try and see if it passes the CI

packages/@stimulus/core/src/application.ts
adrienpoly: will go as suggested with infinite tense

ssaunier:
Thanks @adrienpoly for this work, can't wait to have a logger too! Typos in HTML resulting in not having the Stimulus Controller loaded happen, on dev habbit counter-measure has been `console.log('👋')` in the `connect()`, that change will improve the Developer Experience.

🤔 I wonder about the default though. From your description, it seems that we'd have to manually add that to our pack:

```js
application.debug = process.env.NODE_ENV === 'development'
```

This opt-in approach is good for retro-compatibility, but in the meantime a lot of teams upgrading Stimulus would miss that feature. What would be the sensible default? I'm leaning towards having that line within the library, while still having the possibility to force no debugging in the application code with:

```js
application.debug = false
```

adrienpoly:
> This opt-in approach is good for retro-compatibility, but in the meantime a lot of teams upgrading Stimulus would miss that feature. What would be the sensible default?

```js
process.env.NODE_ENV === 'development'
```

@ssaunier I think the ENV variable work when you have a build system in place such as webpack but would not work if you are using the UMD version of Stimulus or the new autoloading approach from https://github.com/hotwired/stimulus-rails with Skypack.

So unless we would have a per build type default value for the debugger I think we need to default it to false.

db0sch:
btw, that makes me think of something for v2 of this feature: wouldn't it be cool to log when a `data-controller=...` does not match any existing controller? 
It should be a warning (not an error), as this might be intentional, and of course, only in development (and probably upon developer's opt-in, like this logger here).
Would save quite some times. Espacially as I'm using namespaces in my controller's naming, and can be prone to typos and spelling mistakes (convention translation from lower snake case to kebab case... etc)

weaverryan:
This looks incredible! Thank you @adrienpoly (I actually searched for this after seeing your debug mode over on stimulus-use (https://stimulus-use.github.io/stimulus-use/#/docs/debug).

I know from personal experience that this is rather annoying, but does anyone know if there's a plan/timeline to merge some PR's like this and create a 2.1 release? I'm working on a tutorial for Stimulus for the Symfony world and this feature, in particular, would be great to show off.

Also: is there anything we know blocking the PR? Missing tests?

Thanks!

daya:
While we are waiting for the PR to be merged, not knowing what's preventing the merge, I tried the suggestion https://stimulus-use.github.io/stimulus-use/#/docs/debug and couldn't get the `console.log` or `debugger` to work. So presently I don't have a way to debug Stimulus controllers. Any help?

adrienpoly:
@daya StimulusUse debug mode is only for some StimulusUse mixins (by example useDistpactch, useVisibility some others are still in progress). It does noyt cover all of the standard lifecycle events of Stimulus. This PR does

**To the maintainers**
I have seen that this PR https://github.com/hotwired/stimulus/pull/365 has been merged. It does had some test helpers for mocking the logger. This could be use to test this debug feature.

dhh:
This is great @adrienpoly. I'd like to follow this up with a default assignment of the stimulus application instance to window, just like we do with Turbo now, such that it's easy to toggle on the debug mode.

leastbad:
Please reconsider adding Stimulus to the global window by default, as there's no reason to force this on people who don't need/want it.

Anyone that does want it can add
```js
window.Stimulus = Application
```
to their `index.js`.

dhh:
This would make it easier to debug Stimulus both as a way to turn on/off logging, including in production, but also to inspect the state of registered controllers.

Do you have any cases where assigning stimulus to window would cause problems? Of course, even if we did move to doing this by default, one could also just add window.Stimulus = null to clear it out.

leastbad:
I understand the functionality and the mechanics. I also think this PR is a nice enhancement. My comment is specifically in regards to adding global variables that aren't necessary.

I don't have anything to add beyond what I've said. My previous comment stands on its own merits.

A compromise might be to add the assignment to the standard `index.js` template, allowing folks to remove it easily while giving newcomers a globally accessible default.

dhh:
Good point, @leastbad. I've updated both the Stimulus handbook and the default generated index.js in stimulus-rails to use the window.Stimulus assignment. Then it's easy enough for someone to change to use a local variable, if they don't think they benefit from the debugging benefits of the global 👍

adrienpoly:
👍 for merging this PR

about the window assignment I will do more tests but potentially it could cause issues in app where multiple packs are loaded (each pack having Stimulus). With the assignment to `const application` it was handled transparently but as a global variable it probably requires having different names for each pack

dhh:
Yeah, that's why we're better off just doing the default setup as generated code, and then people can change that if they have unique situations like that.

PR #355, opened 2020-12-20
Change the default event for checkbox and radio inputs from "input" back to "change"

jellymann:
https://github.com/stimulusjs/stimulus/commit/14ba2abf75c7ce97e015199996239645d93cd1a9 changed the default event for all `input` elements (except for submit) from `change` to `input`. While this is great for most inputs, radio and checkbox inputs stopped functioning properly, since they do not fire input events at all. This PR aims to resolve this issue.

Resolves: https://github.com/stimulusjs/stimulus/issues/248#issuecomment-748596133

### DIFF
--- a/docs/handbook/03_building_something_real.md
+++ b/docs/handbook/03_building_something_real.md
@@ -93,15 +93,17 @@ We want a click on the button to invoke the `copy()` method in our controller, s
 >
 > Certain other elements have default events, too. Here's the full list:
 >
-> Element           | Default Event
-> ----------------- | -------------
-> a                 | click
-> button            | click
-> form              | submit
-> input             | input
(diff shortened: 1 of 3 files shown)


rik:
That is not the case though. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event#Browser_compatibility) shows that it is working for every current version of browsers. The only browsers where it does not are IE and non-Chromium Edge versions.

jellymann:
Ah, and apparently the version of JSDOM that I'm using, since I picked this up from my tests failing 🤦‍♀️ 

I'll look into this more on my side, feel free to close 👍

javan:
✌️

Issue #363, opened 2021-01-07
Event not bubbling to parent controller


ryanb:
If I have two of the same controllers nested, the action does not bubble up to the parent controller.

```html
<div data-controller="gallery" data-action="click->gallery#next">
  <div data-controller="gallery">
    <button>Click me!</button>
  </div>
</div>
```

Here the `gallery#next` action is not triggered even though the event bubbles up to this DOM element.

Here is a [CodePen example](https://codepen.io/rbates/pen/XWjYGRM).
 
There you can see that if the parent controller is different from the child controller it triggers the action correctly. I expect nesting the same controllers to behave the same way.

This is using Stimulus 2.0.0.

adrienpoly:
I had some other issues with nested controllers couldn't really find a solution. 
I opened a PR #364 with 2 failing tests, yours and mine, this is as far as I could get for now.....

Not sure if they are related.

Hopefully someone can propose a nice solution for both

tdegrunt:
Interestingly I have the reverse experience with a stimulus 1.1.1 controller, whereby an event bubbles even though I ask for it not to do so:

```javascript
export default class extends Controller {
  remove(event) {
    console.log("remove")
    event.cancelBubble = true
    // this.element.remove() - // This causes the event to bubble
    setTimeout(() => {
      this.element.remove() // This setTimeout tricks makes it not bubble
    }, 0)
  }
}
```

Clicking the inner button deletes the outer group, but displays 'remove' twice in the console.

```html
<div data-controller="group">
  <button data-action="rules-group#remove"></button>
  <div data-controller="group">
    <button data-action="rules-group#remove"></button>
  </div>
</div>
```

sstephenson:
Hey Ryan, thanks for opening the issue! This is [intentional behavior](https://discuss.hotwire.dev/t/are-recursive-controllers-explicitly-supported/421), but I admit we've done a poor job communicating it in the documentation.

The main concept to understand is [scopes](https://stimulus.hotwire.dev/reference/controllers#scopes). A scope is the set of elements known to a controller. Specifically, the controller's scope says which elements are eligible to be targets and which elements can connect actions.

In general, a controller's scope is the controller element plus all its children:

<img width="430" alt="Screen Shot 2021-01-09 at 2 00 32 PM" src="https://user-images.githubusercontent.com/2603/104107606-1fa48080-5283-11eb-9fae-5bbcd4226ed6.png">

[...]

Issue #307, opened 2020-05-14
Parent controllers is listening to actions within child controllers


firewalker06:
I have asked in https://discourse.stimulusjs.org/t/is-it-intentional-that-parent-controllers-also-listens-to-actions-within-child-controller/1127 and told to file a bug report here.

---

Hi 👋,

Using example from https://stimulusjs.org/reference/actions#global-events, let say I have the following controllers nested like below:

```html
<div data-controller="gallery">
  <div data-controller="gallery"
       data-action="resize@window->gallery#layout">
    …
  </div>
</div>
```

I found out that the parent `gallery` controller is also listening to the child action. The example above will fires `gallery#layout` twice when the window is resized.

Here is another example in https://codepen.io/javan/pen/JjYMYKO?editors=1010, thanks to @javan 
👍 .

FYI, I’m using Stimulus 1.1.1.

kroehre:
What about using `event.stopPropagation()` in the `layout()` method? I think that would prevent the event from bubbling up to the parent controller. Or, `event.preventDefault()` combined with a defaultPrevented check? 
https://stimulusjs.org/reference/actions#event-objects

BuddyLReno:
Since you declare the controller twice, the event is hooked up twice. `stopPropagation()` won't help in this case because there are two separate event listeners being added to the same event at the same level (window in this case). You will want to create a separate controller for the children from the parents. Event bubbling is just going up the dom tree and checking at each node if there are any listeners for that event.

I don't think this is a bug.

dancallaghan:
This is a bug. 

In the following contrived example, the `next` event should only fire on `.gallery-child` because each controller creates its own scope and the event should be limited to the controller scope it lives under.

```html
<div data-controller="gallery" class="gallery-parent">
  <!-- .gallery-parent scope -->
  <div data-controller="gallery" class="gallery-child">
    <!-- .gallery-child scope -->
    <button type="button" data-action="gallery#next">Next</button>
  </div>
</div>
```

I think the issue might be [`willBeInvokedByEvent`](https://github.com/stimulusjs/stimulus/blob/master/packages/@stimulus/core/src/binding.ts#L63)… need to do some more investigation.

dancallaghan:
So yeah, it looks like the `eventTarget` is `document`/`window` in these cases and it is not correctly scoping the event
https://github.com/stimulusjs/stimulus/blob/master/packages/@stimulus/core/src/binding.ts#L63

javan:
[...]

PR #364, opened 2021-01-07
adds two failing tests for actions

adrienpoly:
I encountered an issue recently with global events and nested controllers https://github.com/hotwired/stimulus/issues/307#issuecomment-749578663

also #363 is another example of actions not being called on the proper controller

I tried a bit to find a solution but without much success so fare. This PR just adds tow failing tests if it can help someone figure out a nice solution


marcoroth:
Yeah, I think it technically still is. 

Unless we just want to say "it's the way how it works and we are not going to change it"

marcoroth:
But I'm going to take a look if I can come up with something useful to solve this and #363.

dhh:
Hmm, I don't think we can/want to mess with the default here. What we could do is add an explicit descriptor somewhere that makes further bubbling up opt-in.

dhh:
But if we go that route, I don't think these tests make sense anyway. Need to make it explicit and opt-in.

Issue #336, opened 2020-09-24
v2 Proposed Enhancement: fooTargetsChanged callback


leastbad:
The ValueChanged callback has been a joy to work with, and it would be amazing to apply a similar mechanism to targets as well.

Having a `fooTargetsChanged()` callback would be good. `fooTargetAdded(element)` and `fooTargetRemoved(element)` would be great. Both would be sublime.

As an example, here is a controller that flashes the last target in a collection when it is inserted with a [CableReady](https://cableready.stimulusreflex.com) [insert_adjacent_html](https://cableready.stimulusreflex.com/usage/dom-operations/element-mutations#insertadjacenthtml) operation:

```js
import { Controller } from 'stimulus'
import { gsap } from 'gsap'
export default class extends Controller {
  static targets = ['location']
  connect () {
    this.element.addEventListener('cable-ready:after-insert-adjacent-html', this.locationInserted)
  }
  disconnect () {
    this.element.removeEventListener('cable-ready:after-insert-adjacent-html', this.locationInserted)
  }
  locationInserted = () => {
    gsap.fromTo(this.locationTargets[this.locationTargets.length - 1], { backgroundColor: "#FF0" }, { backgroundColor: "transparent" })
  }
}
```

If the proposed callback was implemented, this controller could be dramatically simplified:

```js
import { Controller } from 'stimulus'
import { gsap } from 'gsap'
export default class extends Controller {
  static targets = ['location']
  locationTargetAdded (element) {
    gsap.fromTo(element, { backgroundColor: "#FF0" }, { backgroundColor: "transparent" })
  }
}
```

leastbad:
Talk to us, @panoply! I welcome other good ideas. How would you tackle this?

dhh:
Now implemented through #350.

PR #367, opened 2021-01-20
Fire callbacks when targets are added or removed

seanpdoyle:
Fire callbacks when targets are added or removed

Closes [hotwired/stimulus#336][]

---

Implements the `TargetObserver` to monitor when elements declaring
`[data-${identifier}-target]` are added or removed from a `Context`
_after_ being connected and _before_ being disconnected.

In support of iterating through target tokens, export the
`TokenListObserver` module's `parseTokenString` function.

[hotwired/stimulus#336]: https://github.com/hotwired/stimulus/issues/336

### DIFF
--- a/docs/reference/targets.md
+++ b/docs/reference/targets.md
@@ -88,7 +88,7 @@ if (this.hasResultsTarget) {
 }
 ```
 
-## Addition and Removal Callbacks
+## Connected and Disconnected Callbacks
 
 Target _element callbacks_ let you respond whenever a target element is added or
 removed within the controller's element.
--- a/docs/reference/targets.md
+++ b/docs/reference/targets.md
@@ -93,7 +93,7 @@ if (this.hasResultsTarget) {
 Target _element callbacks_ let you respond whenever a target element is added or
 removed within the controller's element.
 
-Define a method `[name]TargetAdded` or `[name]TargetRemoved` in the controller, where `[name]` is the name of the target you want to observe for additions or removals. The method receives the element as the first argument.
+Define a method `[name]TargetConnected` or `[name]TargetDisconnected` in the controller, where `[name]` is the name of the target you want to observe for additions or removals. The method receives the element as the first argument.
 
 Stimulus invokes each element callback any time its target elements are added or removed after `connect()` and before `disconnect()` lifecycle hooks.
 
@@ -101,11 +101,11 @@ Stimulus invokes each element callback any time its target elements are added or
 export default class extends Controller {
   static targets = [ "input" ]
 
-  inputTargetAdded(element) {
+  inputTargetConnected(element) {
     element.classList.add("added-animation")
   }
 
-  inputTargetRemoved(element) {
+  inputTargetDisconnected(element) {
     element.classList.add("removed-animation")
   }
 }
--- a/docs/reference/targets.md
+++ b/docs/reference/targets.md
@@ -99,15 +99,18 @@ Stimulus invokes each element callback any time its target elements are added or
 
 ```js
 export default class extends Controller {
-  static targets = [ "input" ]
+  static targets = [ "item" ]
 
-  inputTargetConnected(element) {
-    element.classList.add("added-animation")
+  itemTargetConnected(element) {
+    this.sortElements(this.itemTargets)
   }
 
-  inputTargetDisconnected(element) {
-    element.classList.add("removed-animation")
+  itemTargetDisconnected(element) {
+    this.sortElements(this.itemTargets)
   }
+
+  // Private
+  sortElements(itemTargets) { /* ... */ }
 }
 ```
 
--- a/docs/reference/targets.md
+++ b/docs/reference/targets.md
@@ -88,6 +88,29 @@ if (this.hasResultsTarget) {
 }
 ```
 
+## Addition and Removal Callbacks
+
+Target _element callbacks_ let you respond whenever a target element is added or
+removed within the controller's element.
+
+Define a method `[name]TargetAdded` or `[name]TargetRemoved` in the controller, where `[name]` is the name of the target you want to observe for additions or removals. The method receives the element as the first argument.
+
+Stimulus invokes each element callback any time its target elements are added or removed after `connect()` and before `disconnect()` lifecycle hooks.
+
+```js
+export default class extends Controller {
+  static targets = [ "input" ]
+
+  inputTargetAdded(element) {
+    element.classList.add("added-animation")
+  }
+
+  inputTargetRemoved(element) {
+    element.classList.add("removed-animation")
+  }
+}
+```
+
 ## Naming Conventions
 
 Always use camelCase to specify target names, since they map directly to properties on your controller.
--- a/packages/@stimulus/core/src/target_observer.ts
+++ b/packages/@stimulus/core/src/target_observer.ts
@@ -1,54 +1,86 @@
+import { Multimap } from "@stimulus/multimap"
+import { Token, TokenListObserver, TokenListObserverDelegate } from "@stimulus/mutation-observers"
 import { Context } from "./context"
-import { Controller } from "./controller"
-import { TokenListObserver, TokenListObserverDelegate, Token, parseTokenString } from "@stimulus/mutation-observers"
+
+export interface TargetObserverDelegate {
+  targetConnected(element: Element, name: string): void
+  targetDisconnected(element: Element, name: string): void
+}
 
 export class TargetObserver implements TokenListObserverDelegate {
   readonly context: Context
-  readonly controller: Controller
-  readonly attributeName: string
-  private tokenListObserver: TokenListObserver
+  readonly delegate: TargetObserverDelegate
+  readonly targetsByName: Multimap<string, Element>
+  private tokenListObserver?: TokenListObserver
 
-  constructor(context: Context, controller: Controller) {
+  constructor(context: Context, delegate: TargetObserverDelegate) {
     this.context = context
-    this.controller = controller
-    this.attributeName = `data-${context.identifier}-target`
-    this.tokenListObserver = new TokenListObserver(this.context.element, this.attributeName, this)
+    this.delegate = delegate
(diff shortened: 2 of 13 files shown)


docs/reference/targets.md
```diff
@@ -88,6 +88,29 @@ if (this.hasResultsTarget) {
...
+
+Define a method `[name]TargetAdded` or `[name]TargetRemoved` in the controller, where `[name]` is the name of the target you want to observe for additions or removals. The method receives the element as the first argument.
+
+Stimulus invokes each change callback any time its target elements are added or removed.
+
+```js
+export default class extends Controller {
+  static targets = [ "input" ]
+
+  inputTargetAdded(element) {
```
seanpdoyle, line 104: @sstephenson to match the Controller's `connect()` and `disconnect()`, would it be better to name these `inputTargetConnect()` and `inputTargetDisconnect()`?

leastbad: My $0.02 is that `added`/`removed` makes more sense because you're addressing elements, not controllers.

leastbad:
I'm extremely happy to see this, @seanpdoyle! You rock.

seanpdoyle:
I've pushed up some changes to the test suite so that it finishes its run. 

It's currently failing, but that's an improvement over the previous version which ran indefinitely due to some infinite recursion.

packages/@stimulus/core/src/target_observer.ts
```diff
@@ -0,0 +1,50 @@
...
+  start() {
+    this.tokenListObserver.start()
+  }
+
+  stop() {
+    this.tokenListObserver.stop()
+  }
+
+  tokenMatched(token: Token): void {
+    if (this.controller.isConnected && this.containsDescendantWithToken(token.element, token.content)) {
```
seanpdoyle, line 27: @sstephenson this PR adds the `isConnected` property to a controller to publish whether or not its `connect()` or `disconnect()` hooks have been invoked.

In this case, it's important to know so that callbacks are not fired during a controller's initial connection via the underlying `TokenListObserver.refresh()` call.

packages/@stimulus/core/src/target_observer.ts
```diff
@@ -0,0 +1,50 @@
...
+      this.dispatchCallback(`${token.content}TargetAdded`, token.element)
+    }
+  }
+
+  tokenUnmatched(token: Token): void {
+    if (this.controller.isConnected && !this.containsDescendantWithToken(token.element, token.content)) {
```
seanpdoyle, line 33: Similarly, this `this.containsDescendantWithToken` is important here because we want to fire removal callbacks when the element itself is removed (i.e. `this.context.element.contains(token.element)` returns `false`), and when the element is _present_, but the token has been removed.

packages/@stimulus/core/src/target_properties.ts
```diff
@@ -33,6 +33,9 @@ function propertiesForTargetDefinition(name: string) {
       get(this: Controller) {
         return this.targets.has(name)
       }
-    }
+    },
+
+    [`${name}TargetAdded`]: (element: Element) => {},
+    [`${name}TargetRemoved`]: (element: Element) => {},
```
seanpdoyle, line 39: @sstephenson I've added these blessings, but they might not be necessary. What can I do to minimize the friction with the class-level property interface work that's in flight?

seanpdoyle:
I'm not sure why, but CI is failing on Windows and `Safari 9.1.2 (Mac OS 10.11.6)`. It passes for me locally in all browsers (including `Safari Version 14.0.3 (16610.4.3.1.7)`)  on macOS.

Is it possible that part of the suite, or any other part of this implementation requires polyfills?

packages/@stimulus/core/src/target_properties.ts
sstephenson: Don't worry about what's in flight. But also, these additions aren't necessary. You're already checking whether the callback exists before you invoke it.

packages/@stimulus/core/src/target_observer.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  private containsDescendantWithToken(element: Element, content: string): boolean {
+    let containsTargetToken = false
+
+    for (const token of parseTokenString(element.getAttribute(this.attributeName) || "", element, this.attributeName)) {
+      containsTargetToken = containsTargetToken || token.content == content
+    }
+
+    return containsTargetToken && this.context.element.contains(element)
```
sstephenson, line 45: This doesn't account for nested scopes:

```html
<div data-controller="a">
  <div data-controller="a">
    <div data-a-target="x">
      <!-- This target does not belong to the outermost controller -->
    </div>
  </div>
</div>
```

docs/reference/targets.md
sstephenson: I wonder if this documentation would be better suited for the Lifecycle Callbacks page instead? And we can briefly mention it with a link on the Targets page.

packages/@stimulus/core/src/target_observer.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  private dispatchCallback(method: string, element: Element) {
+    const callback = (this.controller as any)[method]
+    if (typeof callback == "function") {
+      callback.call(this.controller, element)
```
sstephenson, line 51: We shouldn't be calling methods directly on the controller here. Let's set this up like all the other observers: define a delegate interface, have the observer notify the delegate when targets are added or removed, and make the *delegate* responsible for translating those notifications to our special controller methods.

sstephenson (changes requested):
Implementation-wise, this is looking better since the move to TokenListObserver. I'm not sure I understand the motivation for firing callbacks only on targets that appear after the controller connects, though. That really seems to limit the possible use cases.

My intuition is that target lifecycle callbacks would work similarly to value changed callbacks. So just like we call value changed callbacks for all values present when the controller connects, I'd expect we'd also call the connected callbacks for all targets present when the controller connects. Same goes for disconnected callbacks. 

The specific timeline I'd expect is:

* Controller#initialize()
* Controller#{name}TargetConnected()
* Controller#connect()
* ...
* Controller#{name}TargetDisconnected()
* Controller#disconnect()

I've spiked this out in a separate branch, if you'd like to use it as a reference or a rebase target: https://github.com/hotwired/stimulus/compare/target-lifecycle-callbacks

docs/reference/targets.md
```diff
@@ -88,6 +88,29 @@ if (this.hasResultsTarget) {
...
+```js
+export default class extends Controller {
+  static targets = [ "input" ]
+
+  inputTargetConnected(element) {
+    element.classList.add("added-animation")
+  }
+
+  inputTargetDisconnected(element) {
+    element.classList.add("removed-animation")
```
seanpdoyle, line 109: @sstephenson I now realize that the element will already be out of the DOM, so this contrived example doesn't really illustrate the point.

One use case I had in mind was sorting, would that be a better example? Can you think of a use case where having access to the `Element` instance that was removed would be helpful?

seanpdoyle:
@sstephenson I've incorporated your feedback and pushed up the changes. The suite passes for me locally on evergreen browsers, but is [still failing](https://github.com/hotwired/stimulus/pull/367#issuecomment-808617166) in CI.

Are there polyfills we need to enable in our SauceLabs configuration?

adrienpoly:
@seanpdoyle for the CI I think the issue comes from `isConnected` that is not supported by all browsers. https://caniuse.com/?search=isConnected

This was my first result when searching for a Polyfill
https://gist.github.com/eligrey/f109a6d0bf4efe3461201c3d7b745e8f

you might want to try to add it here to see if it makes the CI happy 
https://github.com/hotwired/stimulus/blob/main/packages/%40stimulus/polyfills/index.js

seanpdoyle:
Thanks for the tip @adrienpoly. The only other spot we're relying on `isConnected` is https://github.com/hotwired/stimulus/blob/21f2ca2c9f6e862aec83b19d433c9660e25536cd/packages/@stimulus/mutation-observers/src/element_observer.ts#L132.

Since the suite isn't failing prior to this change, I wonder if that line is properly covered by our suite. Maybe those tests include a polyfill of sorts?

adrienpoly:
It seems to remove most of the prior issues. Not sure to understand the remaining CI error and yes I think this was a missing polyfill in previous versions.

seanpdoyle:
@adrienpoly thanks for the tips! The suite is now passing.

javan:
> The only other spot we're relying on `isConnected` is … Since the suite isn't failing prior to this change, I wonder if that line is properly covered by our suite.


See https://github.com/hotwired/stimulus/pull/161#issuecomment-393989116 for context.

seanpdoyle:
Thanks for the context @javan. Since this polyfill is only used to pass our suite on CI in a legacy Windows environment, what do you think is the best path forward?

javan:
Totally fine to add a polyfill for `isConnected`.

packages/@stimulus/polyfills/index.js
```diff
@@ -15,3 +15,18 @@ if (typeof SVGElement.prototype.contains != "function") {
     return this === node || this.compareDocumentPosition(node) & Node.DOCUMENT_POSITION_CONTAINED_BY
   }
 }
+
+// From https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected#polyfill
+if (!("isConnected" in Node.prototype)) {
```
seanpdoyle, line 20: @javan this was lifted directly from the [Node.isConnected MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected#polyfill), and is different from your proposed polyfill:

```js
if (!("isConnected" in Node.prototype)) {
  Object.defineProperty(Node.prototype, "isConnected", {
    get() {
      return document.documentElement.contains(this)
    }
  })
}
```

Do we have a preference for which we'd like to ship?

javan: Mine was quick sketch. This one looks solid.

PR #369, opened 2021-01-21
Provide hook for an autoloader.

inopinatus:
Add a hook for an outboard autoloader. Suggested usage like:

```
import { Application } from "stimulus"

const controllerFilename = name => `${name.replace(/--/g, "/").replace(/-/g, "_")}_controller`

const application = new Application()
application.register_autoloader((name, app) => {
  import(controllerFilename(name))
    .then(module => app.register(name, module.default))
    .catch(error => console.log(`Failed to autoload controller: ${name}`, error))
})
application.start()
```

### DIFF
--- a/packages/@stimulus/core/src/application.ts
+++ b/packages/@stimulus/core/src/application.ts
@@ -6,11 +6,14 @@ import { Logger } from "./logger"
 import { Router } from "./router"
 import { Schema, defaultSchema } from "./schema"
 
+type Autoloader = (identifier: string, application: Application) => void
+
 export class Application implements ErrorHandler {
   readonly element: Element
   readonly schema: Schema
   readonly dispatcher: Dispatcher
   readonly router: Router
+  private autoloader?: Autoloader
   logger: Logger = console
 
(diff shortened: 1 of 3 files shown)


packages/@stimulus/core/src/application.ts
```diff
@@ -41,6 +44,14 @@ export class Application implements ErrorHandler {
     this.load({ identifier, controllerConstructor })
   }
 
+  register_autoloader(autoloader: Autoloader) {
```
inopinatus, line 47: I didn't feel this small addition of capability could justify changing the parameters to the Application constructor or the `start` wrapper. Even a class declaration seemed overkill. So I went with a basic registration/callback instead.

inopinatus:
This is an intentionally simple implementation of the basic idea.  It could be extrapolated further e.g. an unregister function; multiple autoloaders; an Autoloader class rather than just a function; more consideration for async behaviour etc - if there's any need/desire to do any of that.

dhh:
This is great @inopinatus. Could you resolve the conflict? Will get it into Stimulus 2.1.0 then.

PR #370, opened 2021-01-21
Use attribute filters

inopinatus:
Possible optimisation, but may need benchmarking and validation prior to merge.


sstephenson:
Thank you for the pull request! But we have no tests and no measurement to show the performance implications. I doubt it will have any measurable impact for most apps.

inopinatus:
@sstephenson That's why I'd left it at "draft" status, until I had the opportunity to supply some.

PR #373, opened 2021-01-27
Add @stimulus/multimap to core deps

drwpow:
## Changes

This is a simple fix: in `@stimulus/core`, it imports `@stimulus/multimap` (such as [router.ts#L5](https://github.com/hotwired/stimulus/blob/main/packages/@stimulus/core/src/router.ts#L5)) but forgets to include it in its own dependencies. This PR just adds that.


sstephenson:
Thank you!

PR #386, opened 2021-03-18
Add handbook section about esbuild-plugin-stimulus

zombiezen:
More info here: https://github.com/zombiezen/esbuild-plugin-stimulus

### DIFF
--- a/docs/handbook/07_installing_stimulus.md
+++ b/docs/handbook/07_installing_stimulus.md
@@ -26,6 +26,42 @@ const context = require.context("./controllers", true, /\.js$/)
 application.load(definitionsFromContext(context))
 ```
 
+## Using esbuild
+
+A community plugin for [esbuild](https://esbuild.github.io/) is available to automatically load controller files from a folder in your app.
+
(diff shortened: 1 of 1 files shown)


sstephenson:
Please add this to the wiki instead: https://github.com/hotwired/stimulus/wiki/Controllers-and-Extensions

Thank you!

zombiezen:
Done. I take it that esbuild is not notable enough at this time to merit mention in the handbook, which is understandable. If that changes down the road, I'd be happy to help with the esbuild support.

FWIW, I did not find the wiki to be super discoverable: there's no direct link from the https://stimulus.hotwire.dev/ site. It's a shame, because it does look like there are some useful links there.

sstephenson:
It's not about notability, we just don't have the expertise or time to maintain instructions for every build system. Hence the wiki! 

You're right that we should add a link to it somewhere on the Stimulus site. Would you mind opening a separate issue on the https://github.com/hotwired/stimulus-site repo?

zombiezen:
Sure! Thanks for the context. 🙂

PR #390, opened 2021-03-26
Add texts on using static get methods for classes, targets and values

leehericks:
This pull request adds explicit examples of using `static get ...` to the definition sections of the targets, values and CSS classes reference pages.

As a person who doesn't work daily with Javascript and uses webpacker-free Rails, I often find myself forgetting the appropriate syntax when checking the reference. Specifically, the only reference to using `static get targets()` is buried at the bottom of the [Handbook](https://stimulus.hotwire.dev/handbook/installing#using-without-a-build-system).

Although not included in this PR, [2: Hello, Stimulus](https://stimulus.hotwire.dev/handbook/hello-stimulus) in the Handbook suggests to "Remix on Glitch", which I tried, and Glitch also doesn't accept the static class properties, so maybe a note on alternate syntax should be added there as well?

### DIFF
--- a/docs/reference/css_classes.md
+++ b/docs/reference/css_classes.md
@@ -38,6 +38,15 @@ export default class extends Controller {
 }
 ```
 
+When using Stimulus without a build system, define CSS classes using `static get classes()` methods instead of `static classes = […]` class properties, which aren't supported natively [yet](https://github.com/tc39/proposal-static-class-features/).
+
+```js
+static get classes() {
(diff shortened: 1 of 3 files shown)


dhh:
From what I can tell, all the evergreen browsers now support static class fields, so we should be good without this, yeah?

leehericks:
Yes I just checked and the new version of Safari (14.1) supports this. It didn't support this at the time I created this pull request.

PR #396, opened 2021-04-26
Utilize Javscript Number to parse 'Number' values

mattpolito:
When processing an incoming empty string against a value property with type Number, the result is NaN. However, if we utilize [Javascript Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) to parse the incoming empty string, we will get a `0`. This is ideal as the default for a Number value property is `0`.

It appears to have browser coverage needed, with no downside that I could find to cover these cases.

### DIFF
--- a/packages/@stimulus/core/src/tests/modules/value_tests.ts
+++ b/packages/@stimulus/core/src/tests/modules/value_tests.ts
@@ -42,6 +42,10 @@ export default class ValueTests extends ControllerTestCase(ValueController) {
     this.controller.numericValue = "garbage" as any
     this.assert.ok(isNaN(this.controller.numericValue))
     this.assert.equal(this.get("numeric-value"), "garbage")
+
+    this.controller.numericValue = "" as any
+    this.assert.equal(this.controller.numericValue, 0)
+    this.assert.equal(this.get("numeric-value"), "")
   }
 
   "test boolean values"() {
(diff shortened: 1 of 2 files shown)


dancallaghan:
It looks like there are other situations where the behaviour of these parsing methods differ e.g.

```javascript
parseFloat('1x')
> 1

Number('1x')
> NaN
```

mattpolito:
@dancallaghan Good catch. However, I'd say the notion of getting a number parsed from a stripped result feels clunky. It's not the result I'd want to have, but maybe this is behavior people are expecting? I know it's something we used to with rails slugging 🤷

javan (approved):

dancallaghan:
@mattpolito yeah, I agree. Just wanted to bring it up because people might be relying on this behaviour.

I just [checked the docs](https://stimulus.hotwire.dev/reference/values#types) and it shows how values are encoded but not decoded (unless I missed it somewhere?). It might be an idea to add an extra column showing `Decoded as…`

dhh:
@dancallaghan If you'd like to update the docs, please do open a PR in https://github.com/hotwired/stimulus-site/ ✌️

dancallaghan:
@dhh, I just read the README after I opened the PR and it said to make the PR here, not on the stimulus-site repo. Whoops

dhh:
Yeah that was actually my bad. I misremembered. We need the changes here. It’s the turbo site that needs them made over there. Really do need to get that updated 😄

Issue #346, opened 2020-12-10
Add previousValue argument to *ValueChanged callback


dancallaghan:
At the moment it looks like only the new value is being passed to the value changed callback. It would be quite helpful to have the previous value passed as the second argument, otherwise you have to get/set the previous value against the controller.

It looks like it's possible by passing the [`attributeOldValue`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserverInit/attributeOldValue) option to the mutation observer.

```javascript
loadingValueChanged(newValue, previousValue) {
```

dancallaghan:
Ah, so it looks like the current implementation is passing `get nameValue` as the first argument which handles the type casting. This problem is not as easy as it appears on the surface.

dhh:
If you find a way to do it, please do open a PR for it 👍

PR #407, opened 2021-06-13
Add `oldValue` argument to `{name}ValueChanged` callback

marcoroth:
This PR adds a second argument to the `{name}ValueChanged` callback which holds the previous value.

Here is a simple example:

```js

  static values = { index: Number }

  connect() {
    this.indexValue = 1
  }

  next() {
    this.indexValue++
  }

  indexValueChanged(value, oldValue) {
    console.log("indexValue", value, oldValue)
  }
}
```

This will be printed if this controller connects and `next()` is called once:
```js
"indexValue" - 0 - undefined    // initialize default value
"indexValue" - 1 - 0            // connect()
"indexValue" - 2 - 1            // next()
```

If you have any feedback please let me know. I'm also happy to add documentation.

Thanks!

Resolves https://github.com/hotwired/stimulus/issues/346

### DIFF
--- a/docs/reference/values.md
+++ b/docs/reference/values.md
@@ -98,7 +98,7 @@ The existential property for a value evaluates to `true` when the associated dat
 
 Value _change callbacks_ let you respond whenever a value's data attribute changes.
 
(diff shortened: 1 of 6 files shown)


dhh:
This is great! Please do add some documentation.

marcoroth:
I've updated the documentation to address the previous values 👍🏼

PR #413, opened 2021-06-20
Emit warnings for undefined controllers, actions and targets

### LINKED ISSUE #324: No error when target or controller is incorrect
I've had multiple issues where I've tried to set a target on a controller only to mistype or get the name of either the controller or the target wrong. Stimulus happily keeps on trucking without complaining when this happens and I have to debug things. It's even more frustrating when I get the controller name wrong on an action - Stimulus doesn't complain that it can't find the controller, it just does nothing.

I've been enjoying using this quite a bit, but better error checking is a must to improve the development experience.

marcoroth:
This PR allows Stimulus to emit warnings if a controller, action or target name is misspelled, incorrect or non-existent.

Previously Stimulus just ignored the wrong declared objects and just did nothing which makes it especially hard for newcomers to grok what's going on. This PR helps in that matter because it emits warnings upfront.

The following scenarios are covered and emit a warning if:

* The declared controller in the `data-controller` attribute doesn't exist or isn't registered in the Application
* The declared method in the `data-action` attribute doesn't exist on the given controller
* The declared controller in the `data-action` attribute doesn't exist
* The declared target in a `data-target` or `data-[identifier]-target` doesn't exist in the `static targets` array of the controller.
* The declared controller in a `data-target` doesn't exist
* The `data-[identifier]-target` attribute is empty
* The format of the `data-target` attribute is wrong or if the value is empty

You can also silence the warnings by setting `warnings` to `false`:

```diff
  const application = Application.start()
  const context = require.context("./controllers", true, /\.js$/)
  application.load(definitionsFromContext(context))
+ application.warnings = false
```

Resolves #324

### DIFF
--- a/packages/@stimulus/core/src/application.ts
+++ b/packages/@stimulus/core/src/application.ts
@@ -13,6 +13,7 @@ export class Application implements ErrorHandler {
   readonly router: Router
   logger: Logger = console
   debug: boolean = false
+  warnings: boolean = true
 
   static start(element?: Element, schema?: Schema): Application {
     const application = new Application(element, schema)
@@ -74,7 +75,9 @@ export class Application implements ErrorHandler {
   // Warning handling
 
   handleWarning(warning: string, message: string, detail: object) {
-    this.logger.warn(`%s\n\n%s\n\n%o`, message, warning, detail)
+    if (this.warnings) {
+      this.logger.warn(`%s\n\n%s\n\n%o`, message, warning, detail)
+    }
   }
 
   // Error handling
--- a/packages/@stimulus/core/src/application.ts
+++ b/packages/@stimulus/core/src/application.ts
@@ -71,6 +71,12 @@ export class Application implements ErrorHandler {
     return context ? context.controller : null
   }
 
+  // Warning handling
+
+  handleWarning(warning: string, message: string, detail: object) {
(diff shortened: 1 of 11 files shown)


existentialmutt:
Good call!  Been using stimulus for years and debugging typoed attributes is still a big time suck.

leastbad:
**Thank goodness.** 👏

dhh:
@marcoroth Trying to use this in anger right now, and I'm seeing that this is causing false warnings for undefined controllers when connecting actions on what appears elements that declare multiple controllers. Seems like when the first controller is detected, the actions on that element are all inspected, and if one of the actions refer to one of the other controllers mention on that element, then you get an error. Haven't fully traced it down yet.

dhh:
Here's a test case:

```html
<html>
<head>
  <script type="module">
    import { Application, Controller } from "https://ga.jspm.io/npm:@hotwired/stimulus@3.0.0-beta.2/dist/stimulus.js"

    application.register("first", class extends Controller {
      hello() {
        alert("hello!")
      }
    })

    application.register("second", class extends Controller {
      goodbye() {
        alert("goodbye!")
      }
    })
  </script>
</head>
<body>
  <div data-controller="first second" data-action="click->first#hello click->second#goodbye">
  </div>
</body>
</html>
```

Produces:
Screen Shot 2021-09-22 at 16 32 00

PR #409, opened 2021-06-14
Fire callbacks when targets are added or removed

Sub-Xaero:
I take no credit for this, Sean Doyle did all the hard work in #367. If Sean no longer wants to commit his time to maintaining this PR, I understand. But it would be a terrible shame for this work to be abandoned at the 11th hour, as this is a piece of functionality I think a lot of Stimulus developers have on their wishlist. 

If all parties are happy for the work he did to still go forward, I'm happy to take this over the line.

PR #415, opened 2021-07-01
Change `String` to `string` for event type

adrienpoly:
this should fix the CI

### DIFF
--- a/packages/@stimulus/core/src/context.ts
+++ b/packages/@stimulus/core/src/context.ts
@@ -86,7 +86,7 @@ export class Context implements ErrorHandler, TargetObserverDelegate {
     return this.element.parentElement
   }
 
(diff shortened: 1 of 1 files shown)


dhh:
Looks like there's still an IE11 error:

Screen Shot 2021-07-01 at 9 16 45 AM

Although I'd also be fine to drop compatibility with IE11.

marcoroth:
The CI started to fail after merging #409. We are probably missing a polyfill for the introduced code.

Sub-Xaero:
I've found the issue, IE11 doesn't like the syntax of the given polyfill. I'll PR a fix.

Sub-Xaero:
C/i passed over on #417 :+1

adrienpoly:
excellent @Sub-Xaero, closing in favor of #417

PR #416, opened 2021-07-01
Explicitly cast the type param in the dispatch method

adammiribyan:
Fixes the TypeScript error which made the test build fail in this PR:
https://github.com/hotwired/stimulus/pull/302


adammiribyan:
@dhh just a tiny change to get the build to pass. I'll work on adding some tests/documentation for the `dispatch` method some time next week.

PR #419, opened 2021-07-05
Fix falsey custom default values

dancallaghan:
When specifying `0` as the default value:

```javascript
static values = { index: { type: Number, default: 0 } }
```

The value getter would incorrectly return the type definition:

```javascript
indexValueChanged() {
  console.log(this.indexValue);
  // => { type: Number, default: 0 }
}
```

This seems like an unlikely scenario but it should still be fixed e.g. people may be specifying a default value via an environment variable.

### DIFF
--- a/packages/@stimulus/core/src/tests/controllers/default_value_controller.ts
+++ b/packages/@stimulus/core/src/tests/controllers/default_value_controller.ts
@@ -13,6 +13,7 @@ export class DefaultValueController extends Controller {
 
     defaultNumber: 0,
     defaultNumberThousand: { type: Number, default: 1000 },
+    defaultNumberZero: { type: Number, default: 0 },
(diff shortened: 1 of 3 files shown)


dancallaghan:
I was just thinking that maybe the function needs some type checking on the return value? i.e. don't return an object for a boolean value type. I'm sure there's a way to do it, but my ts-fu is weak.

marcoroth:
Hey @dancallaghan. Good catch, I missed that edge case. Thank you!

I've put up a PR against your fork (https://github.com/dancallaghan/stimulus/pull/1) to add another test case for a falsey boolean value.

dancallaghan:
Thanks @marcoroth

marcoroth (approved):

PR #423, opened 2021-07-16
Explain how to form identifiers for namespaced controllers in the reference

monorkin:
Hey!
I noticed that the reference is missing an explanation on how to form identifiers for namespaced controllers.

This confused my coworkers and lead them to think that namespacing isn't possible at all, so I though it might be helpful to repeat the instructions from the handbook within the reference's `Identifiers` section.

P.S. I am unsure how `<meta data-controller="callout" data-callout-text-value="articles--reference">` works, and couldn't find a guide how to build the documentation locally to test it, so it's possible I used it incorrectly.

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -45,6 +45,15 @@ For example, this element has a controller which is an instance of the class def
 <div data-controller="reference"></div>
 ```
 
+The following is an example of how Stimulus will generate identifiers for controllers in it's require context:
+
+If your controller file is named… | its identifier will be…
+--------------------------------- | -----------------------
+clipboard_controller.js           | clipboard
+date_picker_controller.js         | date-picker
+users/list_item_controller.js     | users\-\-list-item
+local-time-controller.js          | local-time
+
 ## Scopes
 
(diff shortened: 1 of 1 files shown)


marcoroth:
I think this is the section you were looking for:
https://stimulus.hotwired.dev/handbook/installing#controller-filenames-map-to-identifiers

But maybe it would make sense to also show this table in the section you edited?

monorkin:
>But maybe it would make sense to also show this table in the section you edited?

That is the purpose of this PR, my apologies if that wasn't clear form the PR description.

My coworkers only looked in the reference and expected everything API and naming related to be in it. They expected that the guide would be more of a walk-through.

That this is missing took me by surprise too - I knew how to form the identifier and how to namespace the controller, but couldn't remember where i read it. In the end I had to search the whole documentation for `--` to find where namespacing was explained.

Would you prefer if I copied the table from the guide to this section?

That's what I wanted to do at first, but opted to just add a paragraph and an example as everything else is explained in the `Naming Conventions` section.

dhh:
Yeah, I think reprinting that conversion table would make sense.

monorkin:
I updated the commit with the paragraph and table from my previous comment

PR #432, opened 2021-08-31
Move dispatch helper from context to controller

### LINKED ISSUE #431: dispatch function does not work in Stimulus v3.0.0-beta.1
Thanks again for the great experience.
The dispatch function was a long awaited feature of mine.

However, when I tried to run it, it didn't seem to be working properly.
I looked at the PR and expected it to work with `this.dispatch()`, but it only seems to work with `this.context.dispach`.
Does this behave as expected?
https://github.com/hotwired/stimulus/pull/302

Is there something I am missing?
https://codepen.io/nazomikan/pen/eYRJxyr

adrienpoly:
fix #431

This feature had no tests linked to it, this is only a patch. Once we get the test suite back on track (#421) I can try to add something simple to ensure it remains functional.

### DIFF
--- a/packages/@stimulus/core/src/context.ts
+++ b/packages/@stimulus/core/src/context.ts
@@ -86,13 +86,6 @@ export class Context implements ErrorHandler, TargetObserverDelegate {
     return this.element.parentElement
   }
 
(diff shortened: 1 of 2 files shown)


dhh:
#421 hasn't been merged to main yet, so feel free to get a test added here!

Issue #440, opened 2021-09-03
I want key modifiers for keyboard events.


NakajimaTakuya:
There are many times when we want to support keyboard operations as part of accessibility.

In such cases, you may want to include a filtering process in your code so that it only works with certain keys, as shown below.

```javascript
moveTab(evt) {
  switch(evt.key) {
    case 'ArrowLeft':
      ....
    case 'ArrowRight':
      ...
  }
}
```

This is redundant and can easily lead to deviations from the function name.
(It is tempting to disobey the teachings of Stimulus and name the function onKeydown().)

In particular, `ArrowDown/ArrowUp/ArrowLeft/ArrowRight/Enter/Space/Escape/Home/End` are frequently used keys.

I think it would be ideal if we could use `data-action="keydown.left->tabs#switchLeft"` like in some libraries.
Is there any consideration to implement such a modifier?

marcoroth:
Hey @NakajimaTakuya 

I think the library `stimulus-hotkeys` is exactly what you are looking for. Check out the library here: https://www.npmjs.com/package/stimulus-hotkeys

NakajimaTakuya:
@marcoroth 
Thanks, I'll look into using that.

NakajimaTakuya:
stimulus-hotkeys is a really great experiment!
However, if I'm being honest, I'd like to see official support for it.
The 3rd party library requires a set of element selector and controller names as information, which makes the specification redundant, and it lacks the concept of scope of stimulus actions.
And the way to control events is to use `data-action` and `data-hotkeys... And the fact that there are two ways to control events, `data-action` and `data-hotkeys...` is also complicated.
I think this is unavoidable since it is a 3rd party library.
Is there still room for discussion?

leastbad:
@NakajimaTakuya I appreciate you spending some time going through `stimulus-hotkeys`, even if you make it sound like it was written by an alien. 😉 

All libraries are opinionated. While it's true that my opinions and your opinions may differ for legitimate reasons, it's also possible that I have thought about this problem longer than most people. For example, this isn't even my only Stimulus-based hotkeys library: https://github.com/leastbad/stimulus-shortcut

The main reason Stimulus Hotkeys works the way that it does is because it affords complete flexibility and dynamism: you can push an update to the key bindings from the server and it will take effect immediately.

[...]

PR #442, opened 2021-09-06
Add modifier to filter keyboard events.

NakajimaTakuya:
from https://github.com/hotwired/stimulus/issues/440

There are many cases where we want to support keyboard operations as part of accessibility.

In such cases, determining which key is pressed each time and filtering it in the function may promote a divergence between the function name and the behavior when adhering to the teaching that declarative function names should be given.

```javascript
prevTab(evt) {
  if (evt.key ! == 'ArrowLeft') { return; } // here
  ...
}
````

So, I thought it might be possible to express this filtering on the HTML side to make the function implementation more straightforward.

```html
<button type="button" role="tab" data-action="keydown.left->tabs#prevTab">xxx</button>
//  syntax: `data-action="{event}.{filter_modifier}->{controller}#{action}"`
```

I know that there is a third party library that does this filter on the HTML side, using a different attribute value than data-action.
However, I'm hoping that this can be integrated into data-action as a Stimulus feature to make this simpler to achieve.

```typescript
const keyMappings: { [key: string]: string } = {
  "enter":  "Enter",
  "tab":    "Tab",
  "esc":    "Escape",
  "space":  " ",
  "up":     "ArrowUp",
  "down":   "ArrowDown",
  "left":   "ArrowLeft",
  "right":  "ArrowRight",
  "home":   "Home",
  "end":    "End"
}
```

Here is a mapping of the keys supported by the filter modifier.
(I selected the keys that are often used mainly for accessibility support.)


dhh:
This is very cool, and much simpler than I had imagined. Please do proceed. I'd like to see that in Stimulus ❤️

NakajimaTakuya:
@dhh 
I'm really pleased to see the positive response to this proposal. :heart_eyes:

Let me know what you think.
The modifiers I'm proposing are limited to a few keys.
Do you think it would be better to convert event.key with specific rules (ex.kebab-case) to widen the range of support?
(I assume that special keys such as the spacebar are supported separately.)

dhh:
Haven't forgotten about this. Just want to get 3.0 out the door first, then revisit this for 3.1. Still thinking and weighing the very nice implementation and syntax of this PR against the limited scope.

NakajimaTakuya:
@dhh 
Thank you for telling me your plans. 

Don't worry, I'm in no hurry.
I hope this feature will be included, but just as much I hope that Stimulus will continue to be the most fun for you.
Take your time and think about it. I'll always be here.😄

src/core/action.ts
```diff
@@ -2,6 +2,19 @@ import { ActionDescriptor, parseActionDescriptorString, stringifyEventTarget } f
 import { Token } from "../mutation-observers"
 import { camelize } from "./string_helpers"
 
+const keyMappings: { [key: string]: string } = {
```
lb-, line 5: Could this possibility be put on the Stimulus application? Maybe this way the mapping can be modified on a per application basis this way.

NakajimaTakuya: Like action_event?
https://github.com/hotwired/stimulus/blob/main/src/core/action_event.ts

lb-: Sorry, the comment was not clear.

The use case would be running Stimulus but being able to pass in a custom (or extended) `keyMappings` object when initialising the application (maybe as part of the schema) https://github.com/hotwired/stimulus/blob/main/src/core/schema.ts

```
// src/application.js
import { Application, defaultSchema } from "@hotwired/stimulus"

const customSchema = {
  ...defaultSchema,
  keyMappings: {...defaultSchema.keyMappings, a: "a", w: "w", s: "s", d: "d" },
}

window.Stimulus = Application.start(document.documentElement, customSchema);
```

NakajimaTakuya: Thank you for explaining it all so well! :heart:
I agree with your suggestion.
Give me a minute. :)

NakajimaTakuya:
@lb- 
Sorry to keep you waiting. 
I've tried to accommodate you, but what do you think?

radiantshaw:
@NakajimaTakuya Do you think it would be a good idea to also make the modifier accept an integer value? Like `keyup.13->controller#action`. This would allow the use of any key on the keyboard and it won't be limited to the mapping.

NakajimaTakuya:
@radiantshaw I'm assuming that taking an integer value means using KeyboardEvent.keyCode?
(Sorry if this doesn't mean keyCode.)

I think that's already deprecated by the spec.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Also, with or without it, I'm not sure I'd be motivated to adopt it too aggressively since it depends on the physical keys on the keyboard.

radiantshaw:
@NakajimaTakuya Yeah you're right. Sorry I didn't know about the deprecations. Seems like we're good.

But on a separate note, I was thinking if it's possible to also have syntax for `shiftKey`, `metaKey`, and `ctrlKey`. Something like `keyup.ctrl+a->controller#action` (the syntax is just for illustration purpose), that will check if `ctrlKey` was set to true when `a` was pressed. Do you think it's a good idea to implement this? If yes, then can it be done in the same PR or in a different one?

NakajimaTakuya:
@radiantshaw 
I think that's a good thing.
At first I was making this PR a minimum requirement with the desire to keep it simple and bug free.
When I saw your issue (#530), I changed my mind that this modifier should be chainable with a meta key.
And I was waiting for an opportunity.
I'll try to work on that with this PR.

NakajimaTakuya:
I found one issue as soon as I started my research.
It is the following case.

```html
<td role="gridcell"
  tabindex="0"
  data-action=""
    keydown.home->grid#focusFirstCellInRow
    keydown.ctrl.home->grid#focusFirstCellInGrid
  "
  ...
>
```
If `ctrl+home` is pressed (assuming this element has focus), there are several possible behaviors.
* Both the `focusFirstCellInRow` and `focusFirstCellInGrid` functions work.
* Only the `focusFirstCellInGrid` function works.

Some other libraries seem to default to the former behavior, and then add a .extract modifier to make it behave in the latter way.

[vue's document: .exact Modifier](https://vuejs.org/guide/essentials/event-handling.html#key-modifiers)

That seems somewhat complicated to me.
Therefore, I have a hunch that either this metakey support will not be implemented, or it will default to the latter behavior.

I would like to hear what the owners think.
@dhh what do you think?

lb-:
Reading through this, I think it looks pretty powerful from my perspective and gives some room to add more functionality later. Thanks for taking on some ideas and running with it @NakajimaTakuya

NakajimaTakuya:
@lb- Thanks for the review :)

src/core/action.ts
```diff
@@ -10,24 +11,37 @@ export class Action {
...
+    const eventTarget = this.eventTargetName ? `@${this.eventTargetName}` : ""
+    return `${this.eventName}${eventFilter}${eventTarget}->${this.identifier}#${this.methodName}`
+  }
+
+  isFilterTarget(key: string): boolean {
+    if (!(this.keyFilter && this.keyMappings[this.keyFilter])) {
+      return false;
+    }
+
+    return this.keyMappings[this.keyFilter] !== key
```
imanpalsingh, line 44: Hey ! My first PR review on this repo :D 
Are you only checking if `keyFilter` exists in `keyMappings`? If yes, can this be simplified to

```suggestion
  isFilterTarget(key: string): boolean {
    if(this.keyMappings.hasOwnProperty(this.keyFilter)){
      return this.keyMappings[this.keyFilter] !== key;
    }
    return false;

```

NakajimaTakuya:
@dhh 

I just noticed while trying to resolve the conflict that #569 has been merged, so is this PR not worth it anymore?
At least my goal was achieved, so if it's not a problem I'll close it in a few days.

ex.

```javascript
import {Application, Controller} from "https://cdn.skypack.dev/@hotwired/stimulus@3.1.1";

const app = Application.start()
const mapping = {
  "enter":  "Enter",
  "tab":    "Tab",
  "esc":    "Escape",
  "space":  " ",
  "up":     "ArrowUp",
  "down":   "ArrowDown",
  "left":   "ArrowLeft",
  "right":  "ArrowRight",
  "home":   "Home",
  "end":    "End"
};

Object.entries(mapping).forEach(([modifier, key]) => {
  app.registerActionOption(modifier, ({ event }) => {
    return event instanceof KeyboardEvent && event.key === key;
  });
});

/*
button(data-controller="hoge" data-action="keydown->hoge#notify:space") please push space key
*/

app.register('hoge', class extends Controller {
  notify() {
    alert('hello');
  }
})
```

src/core/action.ts
NakajimaTakuya: Thanks, I'll apply it if they decide they still need this PR. 😄

brunoprietog:
I think it would still be nice to have these default modifiers. 

It had also been suggested to have the open modifier for details element as default with Stimulus.

NakajimaTakuya:
@brunoprietog 
Indeed!
It would make controllers more portable if it were implemented by default.
I will try to create a new PR that enables these by default in a manner consistent with `registerActionOption`.

dhh:
Apologies for the very late review. I continue to love this. I would like to see us ship with a default a-z + 0-9 map, though. And I also think we need to have a syntax for combinations. What part of `keydown.cmd+e->controller#action` did not appeal to you? If we have the default set and we have combinations, then I can see a straight shot to ripping our custom hotkeys controller out of HEY and replacing it with this.

NakajimaTakuya:
Thanks for the reply.
I hadn't heard a response for a while, so I was thinking that you had decided that the content of #567 was sufficient.
The combination syntax is something I want and would like to achieve.
So I am very happy with that response.
I will prepare it.
I will also try to implement the [a-z][0-9] mapping.

docs/reference/actions.md
```diff
@@ -65,11 +65,70 @@ input type=submit | click
...
+const app = Application.start(document.documentElement, customSchema);
+```
+
+If you want to subscribe to a compound filter using a modifier key, you can write it like `shift+a`.
+
+```html
+<div data-action="keydown.shift+a->listbox#selectAll" role="option" tabindex="0">...</div>
+```
+
+The list of supported modifier keys is shown below.
```
lb-, line 120: Is this last section needed? This seems to conflict with the table above.

docs/reference/actions.md
```diff
@@ -65,11 +65,70 @@ input type=submit | click
 select            | change
 textarea          | input
 
+
+
+## KeyboardEvent Filter
+
+You may have Actions that you wish to run only when certain keystrokes are received.
```
lb-, line 72:
```suggestion
There may be cases where [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) Actions should only call the Controller method when certain keystrokes are used.
```

Or

```suggestion
When keystrokes are used for [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) Actions, sometimes they should be filtered by the kind of key that was pressed.
```

lb- (commented):
A few thoughts and comments. 

It may also be good to call out that both keydown, keypress and keyup events can be filtered this way, or essentially all KeyboardEvent events. I tried to add this but not sure if it should be more explicit. 


This is an awesome feature - can't wait!

docs/reference/actions.md
```diff
@@ -65,11 +65,70 @@ input type=submit | click
...
+[0-9]    | [0-9]
+
+If you need to support other keys, you can customize the modifier using custom schema.
+
+```javascript
+import { Application, defaultSchema } from "@hotwired/stimulus"
+
+const customSchema = {
+  ...defaultSchema,
+  keyMappings: {...defaultSchema.keyMappings, "@": "@" },
```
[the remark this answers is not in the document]
NakajimaTakuya: Wow, I forgot about that. Thanks for finding it.

docs/reference/actions.md
NakajimaTakuya: @lb- 
I included the list of modifier keys because I thought it should be explicit.
How does this conflict with the table above?

docs/reference/actions.md
```diff
@@ -65,11 +65,68 @@ input type=submit | click
...
+esc      | Escape
+space    | " "
+up       | ArrowUp
+down     | ArrowDown
+left     | ArrowLeft
+right    | ArrowRight
+home     | Home
+end      | End
+[a-z]    | [a-z]
+[0-9]    | [0-9]
```
lb-, line 98: A suggestion for refinement - using backticks on the code bits.

```markdown
| Filter  | Key Name   |
| ------- | ---------- |
| `enter` | Enter      |
| `tab`   | Tab        |
| `esc`   | Escape     |
| `space` | " "        |
| `up`    | ArrowUp    |
| `down`  | ArrowDown  |
| `left`  | ArrowLeft  |
| `right` | ArrowRight |
| `home`  | Home       |
| `end`   | End        |
| [a-z]   | [a-z]      |
| [0-9]   | [0-9]      |
```

OR - if the side borders is not really what we do (by the looks of things).

```markdown
Filter  | Key Name
------- | ----------
`enter` | Enter
`tab`   | Tab
`esc`   | Escape
`space` | " "
`up`    | ArrowUp
`down`  | ArrowDown
`left`  | ArrowLeft
`right` | ArrowRight
`home`  | Home
`end`   | End
[a-z]   | [a-z
[0-9]   | [0-9
```

docs/reference/actions.md
lb-: sorry, my bad - I was confused, I will suggest a table approach in a min if that is ok.

docs/reference/actions.md
```diff
@@ -65,11 +65,68 @@ input type=submit | click
...
+```html
+<div data-action="keydown.shift+a->listbox#selectAll" role="option" tabindex="0">...</div>
+```
+
+The list of supported modifier keys is shown below.
+
+* meta
+* ctrl
+* alt
+* shift
```
lb-, line 123: Sorted alpha, turned into a table, and notes added.

```suggestion
| Modifier | Notes              |
| -------- | ------------------ |
| `alt`    | `option` on MacOS  |
| `ctrl`   |                    |
| `meta`   | `command` on MacOS |
| `shift`  |                    |
```

OR without the side borders

```suggestion
Modifier | Notes
-------- | -----
`alt`    | `option` on MacOS
`ctrl`   |
`meta`   | `command` on MacOS
`shift`  |
```

docs/reference/actions.md
```diff
@@ -65,11 +65,68 @@ input type=submit | click
...
+[0-9]    | [0-9]
+
+If you need to support other keys, you can customize the modifiers using a custom schema.
+
+```javascript
+import { Application, defaultSchema } from "@hotwired/stimulus"
+
+const customSchema = {
+  ...defaultSchema,
+  keyMappings: {...defaultSchema.keyMappings, "at": "@" },
```
lb-, line 107:
```suggestion
  keyMappings: { ...defaultSchema.keyMappings, at: "@" },
```

docs/reference/actions.md
```diff
@@ -65,11 +65,68 @@ input type=submit | click
...
+}
+
+const app = Application.start(document.documentElement, customSchema);
+```
+If you want to subscribe to a compound filter using a modifier key, you can write it like `shift+a`.
```
lb-, line 112: Suggest a new line after this code snippet.

```suggestion

If you want to subscribe to a compound filter using a modifier key, you can write it like `shift+a`.
```

docs/reference/actions.md
```diff
@@ -65,11 +65,68 @@ input type=submit | click
...
+
+```javascript
+import { Application, defaultSchema } from "@hotwired/stimulus"
+
+const customSchema = {
+  ...defaultSchema,
+  keyMappings: {...defaultSchema.keyMappings, "at": "@" },
+}
+
+const app = Application.start(document.documentElement, customSchema);
```
lb-, line 110: most code snippets in the docs do not use semi-colons for JS - I am a fan of semi-colons personally but probably good to align with the existing code.

```suggestion
const app = Application.start(document.documentElement, customSchema)
```

docs/reference/actions.md
NakajimaTakuya: Certainly it would be easier to use a table structure and include explanations, given the existence of meta keys, etc.
Thanks for your kind review.
It is really helpful.

docs/reference/actions.md
```diff
@@ -65,11 +65,71 @@ input type=submit | click
...
+<div data-action="keydown.ctrl+a->listbox#selectAll" role="option" tabindex="0">...</div>
+```
+
+The list of supported modifier keys is shown below.
+
+| Modifier | Notes              |
+| -------- | ------------------ |
+| `alt`    | `option` on MacOS  |
+| `ctrl`   |                    |
+| `meta`   | `command` on MacOS |
```
lb-, line 125: Ahh. Sorry this may cause confusion this way. Developers may think they need to do `+command`. 

```suggestion
| `meta`   | Command key on MacOS |
```

NakajimaTakuya: i see.

songjiz:
@NakajimaTakuya

I found an issue. This PR break bootstrap events, like `shown.bs.dropdown`.

```javascript
class ZIndexController extends Stimulus.Controller {

  static values = { step: { type: Number, default: 1 } }

  increment(e) {
    const by = e.params.by ? Number(e.params.by) : this.stepValue
    this.value += by
  }

  decrement(e) {
    const by = e.params.by ? Number(e.params.by) : this.stepValue
    this.value -= by
  }

  set value(val) {
    this.element.style.zIndex = Number(val)
  }

  get value() {
    return Number(getComputedStyle(this.element).zIndex)
  }

})

```

```html
<div data-controller="z-index">
   <div class="dropdown">
       <button type="button" aria-expanded="false" data-bs-toggle="dropdown" data-action="shown.bs.dropdown->z-index#increment hidden.bs-dropdown->z-index#decrement">
       <div class="dropdown-menu">
           ...
       </div>
  </div>
</div>
```

`shown` will be parsed as `methodName` of the ActionDescriptor, and `bs.dropdown` will be parsed as `keyFilter` of the ActionDescriptor. then `z-index#increment` method will not be executed

lb-:
@NakajimaTakuya may I suggest you raise a standalone issue for this.

radiantshaw:
@songjiz One solution I can think of is adding Stimulus-specific escape characters. For example, if you want the `shown.bs.dropdown` event to be taken literally, then you can use it as `data-action="shown\.bs\.dropdown->z-index#increment"`. But my hunch is that this might get ugly quickly, so we can also reserve some characters that behave like escape mechanisms. For eg: `data-action="[shown.bs.dropdown]->z-index#increment"` where anything inside the `[]` would be taken literally.

NakajimaTakuya:
@songjiz 
@lb- 
@radiantshaw 
Thanks, I have set up an issue on this.
https://github.com/hotwired/stimulus/issues/612

seb-jean:
@NakajimaTakuya I think `PageUp` and `PageDown` are keys that are also often used primarily for accessibility support

NakajimaTakuya:
@seb-jean 

There are certainly a few widgets that have page up/page down assignments!
I guess "feed", "grid", "treegrid", etc. fall into this category.
I think keyMappings is extensible, so we can add it in userland.

https://stimulus.hotwired.dev/reference/actions#:~:text=If%20you%20need%20to%20support%20other%20keys

seb-jean:
> @seb-jean
> 
> There are certainly a few widgets that have page up/page down assignments! I guess "feed", "grid", "treegrid", etc. fall into this category. I think keyMappings is extensible, so we can add it in userland.
> 
> https://stimulus.hotwired.dev/reference/actions#:~:text=If%20you%20need%20to%20support%20other%20keys

I created PR #677

seb-jean:
I think it would be interesting to add `[a-z]`. I know it already exists but you can only use one character at a time.
This would be used for https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/#kbd2_label and for https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/#kbd_label_listbox.

Current, it would be like this

```html
<button data-action="keydown.a->menu#log keydown.b->menu#log keydown.c->menu#log keydown.d->menu#log ..."></button>
```

I wish I had something like this

```html
<button data-action="keydown.[a-z]->menu#log"></button>
```

seb-jean:
@NakajimaTakuya I tried to implement the filter to include all letters from a to z without having to type 26 actions. Unfortunately, I don't see how to do it. Would you have an idea?

Issue #447, opened 2021-09-21
Add Controller#shouldLoad


dhh:
We use Controller#shouldLoad to prevent certain controllers from being loaded given environmental conditions, like native vs web, working around browser issues, etc. Would be good to get this into Stimulus proper. Here's the dynamic hack version of it I was just using in Basecamp:

```js
application.register = function(identifier, controllerConstructor) {
   if (controllerConstructor.shouldLoad) {
     this.load({ identifier: identifier, controllerConstructor: controllerConstructor })
   }
 }.bind(application)
```

seanpdoyle:
Since controllers already rely on the presence of absence of their identifier in a `data-controller` attribute, could the goals of a `static shouldLoad() { ... }` method be achieved in other ways?

For example, on the server side with something like [token_list](https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-token_list):

```html+erb
<div data-controller="<%= token_list "always", "when-mobile" => request.variant.mobile?, "when-admin" => Current.user.admin? %>"></div>
```

On the client side, conditions could be checked in the [initialize()](https://stimulus.hotwired.dev/reference/lifecycle-callbacks#methods) hook to determine whether or not to remove the identifier:

```javascript
import { controllerIsDisabled } from "../utils"

export default class extends Controller {
  initialize() {
    if (controllerIsDisabled()) {
      const { dataset } = this.element
      dataset.controller = dataset.controller.replaceAll(this.identifier, "")
    }
  }
}
```

On a related note, I'm curious if the presence of an _unloaded_ element could affect Stimulus' controller resolution or other DOM focussed operations (for example, `MutationObserver` instances).

Is there a consideration I'm misunderstanding here?

dhh:
Monkeying with the dom would bust caching, so that's why we do the alternation in the JS.

dhh:
We've been running with this setup since before Turbo was released on HEY. We just had it as part of a webpack helper before. So it never made it upstream.

PR #448, opened 2021-09-22
Add Controller#shouldLoad

dhh:
If you have controllers that are only relevant under certain environmental conditions – like user agent, native, global settings – and caching prevents you from removing them from the DOM, you can use shouldLoad to prevent them from being loaded regardless.


dhh:
Closes #447.

PR #453, opened 2021-09-23
move webpack-helpers back to it's own file

marcoroth:
So they can be impored with `stimulus/webpack-helpers` again

See: https://github.com/hotwired/stimulus/blob/v2.0.0/packages/stimulus/webpack-helpers.js


guillaumebriday-pa:
This is really cool @marcoroth 👍

PR #454, opened 2021-09-23
Improving static values indentation in docs and examples

guillaumebriday:
I found it easier to read.


dhh:
The indention is intentional, since it shows just a fragment, and conveys that there is a class around this. Don't find the extra spacing around values helpful either 😄

PR #455, opened 2021-09-23
Revert "Include webpack helpers in proxy package"

dhh:
Reverts hotwired/stimulus#452

### DIFF
--- a/packages/stimulus/index.js
+++ b/packages/stimulus/index.js
@@ -1,2 +1,1 @@
 export * from "@hotwired/stimulus"
-export * from "@hotwired/stimulus-webpack-helpers"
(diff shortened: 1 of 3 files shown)


leastbad:
Hey, uh... we were kind of using those!

dhh:
It wasn't working correctly. Needed to build for second entry point.

The original helpers are still available in their original location, and we've also republished them under @hotwired/stimulus-webpack-helpers. But until we get a better setup, you'd need to manually add the package.

On top of that, the helper is busted with Webpack v5 (which is used in Webpacker 6). So lots of fun with these damn things 😂

leastbad:
Lots of us still happily chugging with Webpacker 5.4.3, but I hear you (and appreciate the efforts).


Adding `@hotwired/stimulus-webpack-helpers` and then changing your `index.js` to
```js
import { Application } from 'stimulus'
import { definitionsFromContext } from '@hotwired/stimulus-webpack-helpers'
```
does indeed work fine as a workaround, for anyone coming to this PR.

PR #461, opened 2021-09-30
Fix package.json module for stimulus package

cheungpat:
It appears that the `stimulus` package (the one published to npm as `stimulus` rather than `@hotwired/stimulus`, has incorrect `module` defined in the `package.json`. The file `stimulus.esm.js` does not exist in the built package.

This causes problem with other packages that still import `stimulus`.


marcoroth:
This should be fixed once #453 is merged!

cheungpat:
@marcoroth oh! you are right

PR #468, opened 2021-10-07
Fix `stimulus` glue package for cdn use

### LINKED ISSUE #467: Uncaught TypeError: Cannot read properties of undefined (reading 'start')
The package URL [](https://unpkg.com/stimulus/dist/stimulus.umd.js) is not working. It redirects to [](https://unpkg.com/stimulus@3.0.1/dist/stimulus.umd.js). Had to use [](https://unpkg.com/@hotwired/stimulus@3.0.1/dist/stimulus.js) to get it fixed.

marcoroth:
Fixes #467

### DIFF
--- a/packages/stimulus/package.json
+++ b/packages/stimulus/package.json
@@ -9,21 +9,15 @@
     "Javan Makhmali <javan@javan.us>",
     "Sam Stephenson <sstephenson@gmail.com>"
   ],
-  "main": "./dist/stimulus.js",
   "module": "./dist/stimulus.js",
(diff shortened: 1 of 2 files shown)


dhh:
Actually, I don't understand how you prevent double inclusion with this? You're referencing:

```
  "module": "./dist/stimulus.js",
  "main": "./dist/stimulus.umd.js",
```

Those files have Stimulus included in them. So why wouldn't you get double inclusion when you do:

```
import { Application } from "@hotwire/stimulus"
```

?

Issue #35, opened 2018-01-04
Communicating between controllers


vojtad:
I know this isn't exactly an issue, but I think it is a good question. I think of your controllers as controllers for components on a page.
How would you suggest to communicate between more controllers?

I have two components `List` (data rendered in a list) and `ListFilter` (bunch of selects and text inputs). List is loaded using AJAX. It should be reloaded every time `ListFilter` changes. I would like to use separate controllers for `List` and `ListFilter` to keep files small and organized.

So, how to trigger a `List` reload when `ListFilter` changes?

georgeclaghorn:
From a Controller instance, you can call `this.application.getControllerForElementAndIdentifier`:

```js
import { Controller } from "stimulus"

export default class ListFilterController extends Controller {
  // ...

  reloadList() {
    this.listController.reload()
  }

  get listController() {
    return this.application.getControllerForElementAndIdentifier(this.element, "list")
  }
}
```

vojtad:
Thanks! I think this is exactly what I was looking for.

I suppose I should pass the element List component is on as first argument for `getControllerForElementAndIdentifier`, right?

So when a List is next sibling to a ListFilter in the DOM, I should call `getControllerForElementAndIdentifier(this.element.nextElementSibling, "list")` to get the ListController instance, right? Or would you suggest different way to find the List's element?

georgeclaghorn:
I wouldn’t suggest relying on the relative orders of HTML elements.

Prefer targets to access particular elements. Can you move the `data-controller="list-filter"` annotation to a shared parent element and make the list element a target?

vojtad:
Yeah, I can do that. This look like a better way to go, thanks.

If it wouldn't be possible the only way I can think of is to create a parent Controller with targets on both `List` and `ListFilter`. Then I would override the `connect` function of the parent Controller to pass `ListController` instance to `ListFilterController` instance.

However, I am not sure whether `ListController` and `ListFilterController` are already connected when the `connect` function for the parent Controller is called. Do you have any idea about this, please?

domchristie:
FWIW [T3](http://t3js.org) (another JS framework for the HTML you already have ;) has a simple messaging system for communicating between modules (or controllers), keeping them loosely coupled. I wonder if this is something that Stimulus might consider?

The T3 equivalent would look something like:

```js
Box.Application.addModule('list-filter', function (context) {
  return {
    onchange: function () {
[...]

PR #494, opened 2021-11-25
Add part about communicating between controllers

alexander-schranz:
Source: https://github.com/hotwired/stimulus/issues/35#issuecomment-355255358

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -196,3 +196,19 @@ class EffectsController extends Controller {
   }
 }
 ```
+
+## Communicating between Controllers
+
+If you need to communicate or call a method of another controller you can use the
+`getControllerForElementAndIdentifier` method from the application.
+
+```js
+class MyController extends Controller {
+  static targets = [ "other" ]
+
+  copy() {
+    const otherController = this.application.getControllerForElementAndIdentifier(this.otherTarget, 'other')
+    otherController.otherMethod()
+  }
+}
+```
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -197,10 +197,9 @@ class EffectsController extends Controller {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


dhh:
Amended a bit, but good to have in there.

alexander-schranz:
@dhh Thx for updating and merging. Maybe we should before adding also an example how to use events to communicate between controllers. But I'm not sure how that could look like as events are more general and would trigger multiple controllers but maybe you have an a good example which we could add here to make this more clear.

dhh:
There's a full example here: https://github.com/alexander-schranz/stimulus/blob/30818b93050489975947a5f0b3ad25bee5f047c7/docs/reference/controllers.md#cross-controller-coordination-with-events

Or are you looking for something else?

alexander-schranz:
Thx for the fast response. Now I'm understanding that example but still got another question, how is the `clipboard` related here to `effects` need they be registered on the same element to work. Or how I can tell the `clipboard` which specific `effects` controller/element need to be called for that copy element? Possible maybe over some id in the action?

NakajimaTakuya:
@alexander-schranz 
Firing an event via the dispatch method will cause bubbling in this example, as the setting is set to bubbles: true.
https://github.com/hotwired/stimulus/blob/d0aa5d56131c389b63e02815a10c30578607179e/src/core/controller.ts#L64

Therefore, the position of the effect controller can be anywhere in the ancestor.
Also, if you use a global event, you can make it behave in relation to anywhere in the tree.

alexander-schranz:
@NakajimaTakuya Oh okay so they are not directly connected. Just the bubbling over the DOM or if disabled its a global event. Thank you for the clarification.

osadasami:
Check here how to communicate parent to child, child to parent, mixed controllers and distinct controllers https://github.com/osadasami/code/issues/1

PR #520, opened 2022-02-21
Docs - add details for `dispatch` options

### LINKED ISSUE #510: document additional options for this.dispatch
I have noticed in the code that `this.dispatch` on a controller allows extra options, specifically a `target` for dispatching an event on a custom target. 

https://github.com/hotwired/stimulus/blob/53189b473651e7288845af88fef661be69f27345/src/core/controller.ts#L64

However, this is not listed in the official documentation

https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events

Proposed solution; document the `target` option and other options supported; `prefix`, `cancelable` and `bubbles`. 

I understand if prefix is something that should not be documented as the consistent event naming is one of the good things about using `this.dispatch`. 

I would be happy to submit a docs PR if this is something useful.

lb-:
* Add documentation for all the options that can be passed to the controller's `this.dispatch` method.
* Add documentation for what `this.dispatch` returns and an example of how to use this to provide cancellable functionality
* Resolves #510

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -215,7 +215,8 @@ class ClipboardController extends Controller {
 
   copy() {
     const event = this.dispatch("copy", { cancelable: true })
-    if (event.defaultPrevented) return;
+    if (event.defaultPrevented) return
+
     this.sourceTarget.select()
     document.execCommand("copy")
   }
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -203,7 +203,7 @@ option       | default            | notes
 -------------|--------------------|----------------------------------------------------------------------------------------------
 `detail`     | `{}` empty object  | See [CustomEvent.detail](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail)
 `target`     | `this.element`     | See [Event.target](https://developer.mozilla.org/en-US/docs/Web/API/Event/target)
-`prefix`     | `this.identifier`  | If prefix is falsey (e.g. `null`), only the `eventName` will be used
+`prefix`     | `this.identifier`  | If the prefix is falsey (e.g. `null` or `false`), only the `eventName` will be used. If you provide a string value the `eventName` will be prepended with the provided string and a colon. 
 `bubbles`    | `true`             | See [Event.bubbles](https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles)
 `cancelable` | `true`             | See [Event.cancelable](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable)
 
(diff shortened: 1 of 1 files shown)


docs/reference/controllers.md
```diff
@@ -197,6 +197,40 @@ class EffectsController extends Controller {
...
 
+`dispatch` accepts additional options as the second parameter as follows:
+
+option       | default            | notes
+-------------|--------------------|----------------------------------------------------------------------------------------------
+`detail`     | `{}` empty object  | See [CustomEvent.detail](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail)
+`target`     | `this.element`     | See [Event.target](https://developer.mozilla.org/en-US/docs/Web/API/Event/target)
+`prefix`     | `this.identifier`  | If prefix is falsey (e.g. `null`), only the `eventName` will be used
+`bubbles`    | `true`             | See [Event.bubbles](https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles)
+`cancelable` | `true`             | see [Event.cancelable](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable)
```
marcoroth, line 208:
```suggestion
`cancelable` | `true`             | See [Event.cancelable](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable)
```

docs/reference/controllers.md
```diff
@@ -197,6 +197,40 @@ class EffectsController extends Controller {
 }
 ```
 
+`dispatch` accepts additional options as the second parameter as follows:
+
+option       | default            | notes
+-------------|--------------------|----------------------------------------------------------------------------------------------
+`detail`     | `{}` empty object  | See [CustomEvent.detail](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail)
+`target`     | `this.element`     | See [Event.target](https://developer.mozilla.org/en-US/docs/Web/API/Event/target)
+`prefix`     | `this.identifier`  | If prefix is falsey (e.g. `null`), only the `eventName` will be used
```
marcoroth, line 206: I think it also makes sense to describe the truthy option

```suggestion
`prefix`     | `this.identifier`  | If the prefix is falsey (e.g. `null` or `false`), only the `eventName` will be used. If you provide a string value the `eventName` will be prepended with the provided string and a colon. 
```

docs/reference/controllers.md
```diff
@@ -197,6 +197,40 @@ class EffectsController extends Controller {
...
+
+`dispatch` will return the generated [`CustomEvent`](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent), you can use this to provide a way for the event to be cancelled by any other listeners as follows:
+
+```js
+class ClipboardController extends Controller {
+  static targets = [ "source" ]
+
+  copy() {
+    const event = this.dispatch("copy", { cancelable: true })
+    if (event.defaultPrevented) return;
```
marcoroth, line 218: I just looked at the code snippets in the docs, it seems like we are not using semi-colons.

```suggestion
    if (event.defaultPrevented) return

```

marcoroth (commented):
Hey @lb-, thank you for opening this pull request!

I just had some minor nit-picks for consistency. But this looks good to me otherwise!

lb-:
Thanks for the suggestions - great feedback @marcoroth 
I have just committed them on GitHub - can squash/rebase next week if needed.

marcoroth:
All good, I will squash merge them! Thank you! 🚀

lb-:
Awesome. Thanks.

PR #528, opened 2022-03-15
Improve Stimulus Values type mismatch error messages

lamartire:
The current implementation isn't friendly for debugging, especially when you have multiple controllers on the same page. If your value doesn't match to the expected type you will see a `Uncaught Error: Type "string" must match the type of the default value. Given default value: "undefined" as "undefined"` message.

This PR changes the message and pass value's name into it. Result will look like `The specified default value for the Stimulus Value "property_name" must match the defined type "string". The provided default value of "123" is of type "number".` and so on.

### DIFF
--- a/src/core/value_observer.ts
+++ b/src/core/value_observer.ts
@@ -85,14 +85,21 @@ export class ValueObserver implements StringMapObserverDelegate {
 
     if (typeof changedMethod == "function") {
       const descriptor = this.valueDescriptorNameMap[name]
-      const value = descriptor.reader(rawValue)
-      let oldValue = rawOldValue
 
-      if (rawOldValue) {
-        oldValue = descriptor.reader(rawOldValue)
+      try {
+        const value = descriptor.reader(rawValue)
+        let oldValue = rawOldValue
+
+        if (rawOldValue) {
+          oldValue = descriptor.reader(rawOldValue)
+        }
+
+        changedMethod.call(this.receiver, value, oldValue)
+      } catch (error) {
+        if (!(error instanceof TypeError)) throw error
+
+        throw new TypeError(`Stimulus Value "${this.context.identifier}.${descriptor.name}" - ${error.message}`)
       }
-
-      changedMethod.call(this.receiver, value, oldValue)
     }
   }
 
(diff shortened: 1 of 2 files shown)


lamartire:
@sstephenson could you please provide review here?

marcoroth (commented):
Hey @lamartire, thank you for opening this pull request! 

I agree that my first pass on this message is not very intuitive, especially at large. I really like where you are going with this!

I'm wondering if it would even make sense to to add the controller identifier to the message. Maybe we can even improve the copy of the message as well. I was thinking something along those lines:

```
Uncaught Error: The specified default value for the Stimulus value "identifier.property_name" must match the type "string". The provided value of "undefined" is of type "undefined".
```

or:
```
Uncaught Error: The specified default value for the Stimulus value "property_name" in controller "identifier" must match the type "string". The provided value of "undefined" is of type "undefined".
```

What do you think?

lamartire:
Yeah, I agree with you. Nice idea!

lamartire:
So, I added it, but I think I need add additional tests for it. Just validate, that everything matches with your vision, then I finalize the PR with specs

lamartire:
@marcoroth did it, but due the blessing definition (static field) we can't get controller when initially – when we're checking default value type. I'm not sure, that it worth to rewrite blessings, let's leave the solution as is.

marcoroth:
@lamartire Thank you for finishing this up! 

Ah yes, I didn't realize that this is happening during the "setup" of the controller instance and thus we don't have access to the identifier yet. 

But as you said, I agree that we shouldn't mess with the blessings just for having the controller name in the error message. There would be some ways to achieve that, but I don't think it's worth it.

I slightly tweaked your PR by removing a redundant if-check ([`19c6894`](https://github.com/hotwired/stimulus/pull/528/commits/19c6894d3867e52be33a235fa1c5433a3f31eb91)) and reworded the error message a bit to match the message in the PR ([`f23f820`](https://github.com/hotwired/stimulus/pull/528/commits/f23f82021a8dfb60392c5ebc4cb7f511da87d1c1)). 

I also updated another error type message which I wanted to tackle ([`af9f834`](https://github.com/hotwired/stimulus/pull/528/commits/af9f834c2b918397e51ea1d7a9997edfb039559b)). Now you also see an updated error message, when setting or overriding a Stimulus Value and the types don't match:

**Before:**
- `"TypeError: Expected object"`
- `"TypeError: Expected array"`

**After:**
- `TypeError: Stimulus Value "my-controller.myObjectValue" -
expected value of type "object" but instead got value "false" of type
"boolean"`

- `TypeError: Stimulus Value "my-controller.myArrayValue" -
expected value of type "array" but instead got value "123" of type
"number"`

marcoroth (approved):

PR #529, opened 2022-03-19
Allow refining the type of Controller elements

rik:
Use generics to avoid having to repeat type assertions.

This can be used as is:
```ts
class MyFormController extends Controller<HTMLFormElement> {
    submit() {
        new FormData(this.element)
    }
}
```

The default value for that generics means you can still write `class MyController extends Controller {}`.

---

This was inspired by [this conversation](https://discuss.hotwired.dev/t/stimulus-and-typescript/2458/3).

### DIFF
--- a/src/core/controller.ts
+++ b/src/core/controller.ts
@@ -6,7 +6,7 @@ import { ValuePropertiesBlessing, ValueDefinitionMap } from "./value_properties"
 
 export type ControllerConstructor = Constructor<Controller>
 
-export class Controller {
+export class Controller<ElementType extends Element = Element> {
   static blessings = [ ClassPropertiesBlessing, TargetPropertiesBlessing, ValuePropertiesBlessing ]
   static targets: string[] = []
   static values: ValueDefinitionMap = {}
@@ -30,7 +30,7 @@ export class Controller {
   }
 
   get element() {
-    return this.scope.element
+    return this.scope.element as ElementType
   }
 
   get identifier() {


marcoroth:
Hey @rik, thank you for opening this pull request!

I agree that it makes sense that you can refine the type of the controller element with generics in TypeScript. The use of the default value also seems super straight-forward, I like that!

Apart from your code changes, I think it would make senes that we add a chapter to the docs which outline some details on how you can use/integrate Stimulus with TypeScript. I've seen a few snippets in the past which might be worth to collect in a dedicated chapter in the docs, including this change here.

Would you mind starting such a chapter in this PR for this feature, so that we have a foundation to build on for other TypeScript related docs snippets in the future?

alexander-schranz:
Maybe I can help here out:


~~~markdown
# Using Typescript

Stimulus itself written is TypeScript and provides types directly over its package.
The following documentation should make it clear how to define types for stimulus
properties.

## Defining Controller Element Types

By default the `element` of the controller is from type `Element`.  
If the element is from another type you can define it over the controller definition this way:

```ts
import { Controller } from '@hotwired/stimulus';

export default class MyController extends Controller<HTMLFormElement> {
    submit() {
        new FormData(this.element)
    }
}
```

## Defining Controller Value Types

To define the type of an configured values is possible
over the declare feature of TypeScript:

```ts
import { Controller } from '@hotwired/stimulus';

export default class MyController extends Controller<HTMLFormElement> {
    static values = {
        code: String,
    };

    declare readonly codeValue: string;
}
```

> The `declare` avoids override of the exist stimulus property and just define the type for TypeScript.

## Defining Controller Target Types

To define the type of configured targets the controller use the
declare feature of TypeScript:

```ts
import { Controller } from '@hotwired/stimulus';

export default class MyController extends Controller<HTMLFormElement> {
    static targets = [
        'name'
    ];

    declare readonly nameTarget: HTMLInputElement;
}
```

> The `declare` avoids override of the exist stimulus property and just define the type for TypeScript.

## Custom properties

Other custom properties is possible to define the normal TypeScript way on the controller:

```ts
import { Controller } from '@hotwired/stimulus';

export default class MyController extends Controller<HTMLFormElement> {
    input: HTMLInputElement;
}
```

Read more about it in the [TypeScript Documentation](https://www.typescriptlang.org/docs/handbook/intro.html).
~~~

marcoroth (approved):
Thanks @rik, this is good as is.

Also thank you to you @alexander-schranz. Would you mind posting this as a separate PR so you can get the proper credit for it? Thank you!

alexander-schranz:
@marcoroth https://github.com/hotwired/stimulus/pull/540 here we go :)

ReazerDev:
Hi, any ETA for when this will be released?
The last update of the npm package was 7 months ago ^^

lb-:
Running into this issue also - looking forward to seeing this released 👍

alexander-schranz:
I don't know when this is going to release. But currently you could add a abstract controller in your project which you extend like here: https://discuss.hotwired.dev/t/stimulus-and-typescript/2458/3 and when released you can remove  it and replace it with the stimulus controller again.

rik:
As far as I can tell, `Element` is equivalent to `HTMLElement | SVGElement` because those are the only two subclasses of Element.

lb-:
> As far as I can tell, `Element` is equivalent to `HTMLElement | SVGElement` because those are the only two subclasses of Element.

But then I run into this. https://stackoverflow.com/a/58773720/8070948

Anyway. Probably de-railing this issue's chat - I'll just ask on the community thread if I get more stuck here.

Issue #530, opened 2022-03-30
Support for some other Action Options and Key Modifiers would be great


radiantshaw:
This issue is inspired by #440, #442, and [Vue's `v-on` directive](https://vuejs.org/api/built-in-directives.html#v-on). Out of the listed "Modifiers", Stimulus already has the following:

- `:capture`
- `:once`
- `:passive`
- `:!passive`

The `.{keyAlias}` Key Modifier is already being handled in #442 for keyboard events. We can have the equivalent `.left`, `.right`, and `.middle` Key Modifiers for mouse events as well.

On top of that, seems like having a `:self` Action Option would be great. Maybe an equivalent `:!self` as well?

And I know that the `:stop` and `:prevent` Action Options aren't necessary since the same can be done inside the handler itself, but it'd be nice to have those options as well to reduce `event.stopPropagation()` and `event.preventDefault()` inside most handlers.

Thoughts?

dhh:
Yeah, I fundamentally like this syntax, and would happy to see it extended to mouse clicks as well. Do investigate.

radiantshaw:
@dhh Sure. I'll take this up and raise an appropriate PR for it. Thanks! 🤝

adrienpoly:
I like this idea for `:stop` `:prevent`

I suppose we would be able to chain them ?
```html
<div data-action="click->controller#vote:once:prevent">
```

dhh:
Ahh, I had thought of click.once->controller, but what you're proposing is even better, @adrienpoly. And yeah, the chaining seems very reasonable. Should still combine this with @NakajimaTakuya's work on the hotkeys and mouse modifiers, so we could have `click.left->controller#vote:once`.

adrienpoly:
yes `click.once->controller` was this initial syntax when we looked into this action options feature (https://github.com/hotwired/stimulus/pull/232) but we ended up with `click->controller#action:once` 

So we would :
- describe event **options** with trailing `:option` tokens
- **filter** events with `.variation` token (probably only one in this case)

## Examples

```html
# mixing event variant and event options
<div data-action="click.left->controller#vote:once:prevent">

# with global listener
<div data-action="keydown.left@window->tabs#prevTab">

```

radiantshaw:
Just an update. I'm also implementing the `:self` functionality in a [separate branch](https://github.com/radiantshaw/stimulus/tree/self-and-not-self-action-options) in my fork. But I'm waiting for #535 to get approved and merged first.

PR #533, opened 2022-04-07
log_controller_test_case: Fix deepEqual returning true for primitive types

### LINKED ISSUE #532: `deepEqual` always returns `true` when comparing two booleans
The `deepEqual` function inside the `src/tests/cases/log_controller_test_case.ts` code:

https://github.com/hotwired/stimulus/blob/b459c94c65b9b3ec468bf5a8b9cb8dcfc63ed8a6/src/tests/cases/log_controller_test_case.ts#L37-L49

always returns `true` when two boolean values are compared.

jcpascual:
This PR fixes ``deepEqual`` to return the correct value when comparing primitive types.

When comparing ``obj1`` and ``obj2``, ``deepEqual`` first performs a standard ``===`` check. If that fails, it then proceeds to do a "deep comparison" with each targets' properties. However, since ``Object.keys(obj)`` always returns an empty array for primitive types, this means that ``deepEqual`` will always return ``true`` when comparing primitives as there are no properties to compare. (For example, ``deepEqual(true, false)`` will result in ``true``.)

To avoid this, we only perform the deep comparison if both targets are of type ``object``. If they are not, we return ``false``.

Fixes #532.

### DIFF
--- a/src/tests/cases/log_controller_test_case.ts
+++ b/src/tests/cases/log_controller_test_case.ts
@@ -37,7 +37,7 @@ function slice(object: any, keys: string[]): any {
 function deepEqual(obj1: any, obj2: any): boolean {
   if (obj1 === obj2) {
     return true
-  } else {
+  } else if (typeof obj1 === "object" && typeof obj2 === "object") {
     if (Object.keys(obj1).length !== Object.keys(obj2).length) { return false }
     for (var prop in obj1) {
       if (!deepEqual(obj1[prop], obj2[prop])) {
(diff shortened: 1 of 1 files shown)


radiantshaw (approved):

radiantshaw:
@dhh If this PR looks good to you, can we please fast-forward merging this? Cause I'll need this for the PR I'm working on for resolving #530.

marcoroth (approved):
Thank you @jcpascual and @radiantshaw! 

This looks good to me, let's get it merged!

PR #535, opened 2022-04-12
Adds new Action Options, namely `:stop` and `:prevent`

radiantshaw:
This PR adds two new Action Options, namely `:stop` and `:prevent`, as discussed in #530. For more context, please refer to the original issue.

Please do not hesitate to give any suggestions and I'll make sure to accommodate those.

**NOTE**: For some of the other things discussed in #530, I'll do them in a separate PR.

### DIFF
--- a/src/core/action_descriptor.ts
+++ b/src/core/action_descriptor.ts
@@ -1,8 +1,8 @@
-import { ExtendedAddEventListenerOptions } from "../extensions"
+import { EventModifiers } from "./event_modifiers"
 
 export interface ActionDescriptor {
   eventTarget: EventTarget
-  eventOptions: ExtendedAddEventListenerOptions
+  eventOptions: EventModifiers
   eventName: string
   identifier: string
   methodName: string
@@ -31,7 +31,7 @@ function parseEventTarget(eventTargetName: string): EventTarget | undefined {
   }
 }
 
-function parseEventOptions(eventOptions: string): ExtendedAddEventListenerOptions {
+function parseEventOptions(eventOptions: string): EventModifiers {
   return eventOptions.split(":").reduce((options, token) =>
     Object.assign(options, { [token.replace(/^!/, "")]: !/^!/.test(token) })
   , {})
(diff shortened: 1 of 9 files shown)


src/core/action_descriptor.ts
```diff
@@ -1,6 +1,8 @@
+import { ExtendedAddEventListenerOptions } from "../extensions"
+
 export interface ActionDescriptor {
   eventTarget: EventTarget
-  eventOptions: AddEventListenerOptions
+  eventOptions: ExtendedAddEventListenerOptions
```
adrienpoly, line 5: One suggestion as we now have properties that are not anymore linked to the EventListner maybe we could name that concept `EventModifiers`

That would be also inlined with the naming conventions in Vue 3 https://vuejs.org/guide/essentials/event-handling.html#event-modifiers

radiantshaw: @adrienpoly Resolved in ec703c4e359faa51313441092bd001bfbb632cbb. 🚀

src/tests/modules/core/event_options_tests.ts
```diff
@@ -131,10 +131,73 @@ export default class EventOptionsTests extends LogControllerTestCase {
...
+    )
+  }
+
   set actionValue(value: string) {
     this.buttonElement.setAttribute("data-action", value)
   }
 
+  set elementActionValue(value: string) {
+    this.element.setAttribute("data-action", value)
+  }
```
radiantshaw, line 199: I don't like this. Maybe we can do something like:

```ts
set actionValueFor(element: HTMLElement, value: string) {
  element.setAttribute("data-action", value)
}
```

And then when calling, we can do:

```ts
this.actionValueFor(this.element, "controller#action")
this.actionValueFor(this.buttonElement, "other-controller#otherAction")
```

But this would require changes to the entire test file. Thoughts?

---

Another option could be doing:

```ts
set actionValue(value: string, config?: { for: HTMLElement }) {
  (config?.for || this.element).setAttribute("data-action", value)
}
```

dhh: Can improve this in a separate PR 👍

radiantshaw: @dhh Sure. I'll raise a minor PR for this. 👍

radiantshaw: #541 resolves this.

PR #539, opened 2022-04-26
Remove obsolete document.execCommand

qwerjulia:
document.execCommand("copy") is obsolete. Although it may still work in some browsers, its use is discouraged since it could be removed at any time.

### DIFF
--- a/examples/controllers/clipboard_controller.js
+++ b/examples/controllers/clipboard_controller.js
@@ -11,7 +11,6 @@ export default class extends Controller {
   }
 
   copy() {
-    this.sourceTarget.select()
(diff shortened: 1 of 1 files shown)


adrienpoly:
Thanks @qwerjulia this indeed would indeed align the example with the update of the documentation made in #484.

However I see a few differences :
- in #484 in connect `if (document.queryCommandSupported("copy")) {` is also replaced
- in #484 is `navigator.clipboard.writeText(this.sourceTarget.value)` whereas you don't add value. Can we have the same for both if you could test a real example ?

adrienpoly:
fixed by #559

PR #541, opened 2022-05-02
Refactor Event Options test helpers

radiantshaw:
Refactor the following lines of code:

https://github.com/hotwired/stimulus/blob/30933d6dceeae47ef19662017fa439805c6bd862/src/tests/modules/core/event_options_tests.ts#L193-L199

and replace it with a new method called `setAction` that accepts an `Element` onto which to set the action. This is similar in signature to `triggerEvent`.

Resolves the conversation: https://github.com/hotwired/stimulus/pull/535#discussion_r855772092.


radiantshaw:
@dhh Can we merge this PR if everything looks okay? Is there anything else that needs to be done?

marcoroth (approved):
Thank you @radiantshaw, I think this looks good!

PR #546, opened 2022-05-29
Adds new Action Options, namely `:self` and `:!self`

radiantshaw:
This PR adds two new Action Options, namely `:self` and `:!self`, as discussed in #530. For more context, please refer to the original issue.

Feel free to provide any suggestions and I'll make sure to accommodate them.

### DIFF
--- a/docs/reference/actions.md
+++ b/docs/reference/actions.md
@@ -107,6 +107,8 @@ Custom action option | Description
 -------------------- | -----------
 `:stop`              | calls `.stopPropagation()` on the event before invoking the method
 `:prevent`           | calls `.preventDefault()` on the event before invoking the method
+`:self`              | only invokes the method if the event was fired by the element itself
+`:!self`             | only invokes the method if the event was fired by any of its descendants but not by the element
 
 ## Event Objects
 
--- a/docs/reference/actions.md
+++ b/docs/reference/actions.md
@@ -108,7 +108,6 @@ Custom action option | Description
 `:stop`              | calls `.stopPropagation()` on the event before invoking the method
 `:prevent`           | calls `.preventDefault()` on the event before invoking the method
 `:self`              | only invokes the method if the event was fired by the element itself
-`:!self`             | only invokes the method if the event was fired by any of its descendants but not by the element
 
 ## Event Objects
 
--- a/src/core/binding.ts
+++ b/src/core/binding.ts
@@ -31,7 +31,7 @@ export class Binding {
   }
 
   handleEvent(event: Event) {
-    if (this.willBeInvokedByEvent(event)) {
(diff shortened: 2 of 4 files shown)


radiantshaw:
Seems like some of the older commits are also showing up here. I'll rebase with the latest main branch and re-push when I get the time. 👍

radiantshaw:
Done. I've rebased the branch with the latest main. 🚀

rik:
I quite liked the clear distinction proposed in https://github.com/hotwired/stimulus/issues/530#issuecomment-1082999316:
> - describe event options with trailing :option tokens
> - filter events with .variation token (probably only one in this case)

I think it will also read better: `<button data-action="click.self->hello#greet">` vs `<button data-action="click->hello#greet:self">`

---

Is there a use case for `!self` ? I can't think of one.

radiantshaw:
@rik Sorry for replying late on this. I've been trying to also think of some use case for `!self` but can't. I'll think of it for some more time and get back on this.

dhh:
Let's just go with `self` alone?

radiantshaw:
@dhh Yes. I didn't find any use case for the `:!self` option. It also makes sense for it to be `.self` instead of `:self` as @rik mentioned. So I'll try to refactor the code.

I was a bit busy so I didn't get much time to work on this. But I do plan on working on it in the upcoming weekends. ✌️

marcoroth:
I think we should keep it as `:self` so it's in line with all the other event modifiers

radiantshaw:
@marcoroth I'm okay with either. But I also agree with @rik and feel that `self` falls into the "filter" category instead of "event options" and it would be better if we keep it as `.self` instead of `:self`. But I'm open to feedback on why we should have it as `:self` as well.

dhh:
Think we want to stick with `:self` to get chaining. You could want both self + prevent etc.

rik:
> Think we want to stick with `:self` to get chaining. You could want both self + prevent etc.

I don't understand this point. The syntax proposed by @adrienpoly allows for self + prevent. [You left a rocket emoji reaction to this proposal](https://github.com/hotwired/stimulus/issues/530#issuecomment-1082999316). I'm copying the examples here:

```html
# mixing event variant and event options
<div data-action="click.left->controller#vote:once:prevent">

# with global listener
<div data-action="keydown.left@window->tabs#prevTab">

dhh:
@rik Sorry, I meant mixing things like left and self. IMO, self feels more like once/prevent than it does "left". Don't like the feel of "click.left.self->" either, even if that could be possible.

seanpdoyle:
I'm sorry to chime in so late, but I'm at a loss understanding `:self`'s utility. I read through the original issue, but couldn't find an example use case. Could you share one here?

Are `:self`'s semantics similar to checking an event's `currentTarget` as it bubbles up the DOM? Was `:currentTarget` ever considered as a modifier name?

PR #567, opened 2022-07-28
Support custom Action Options

seanpdoyle:
As a follow-up to [hotwired/stimulus#535][] and
[hotwired/stimulus#546][], add support for declaring custom action
modifiers in the same style as `:prevent`, `:stop`, and `:self`.

Take, for example, the [toggle][] event. It's dispatched whenever a
`<details>` element toggles either open or closed. If an application
were able to declare a custom `open` modifier, it could choose to route
`toggle` events denoted with `:open` _only_ when the `<details open>`.
Inversely, they could choose to route `toggle` events denoted with
`:!open` _only_ when the `<details>` does not have `[open]`.

Similarly, the same kind of customization could apply to custom events.
For example, the [turbo:submit-end][turbo-events] fires after a `<form>`
element submits, but does not distinguish between success or failure. A
`:success` modifier could skip events with an unsuccessful HTTP response
code.

[hotwired/stimulus#535]: https://github.com/hotwired/stimulus/pull/535
[hotwired/stimulus#546]: https://github.com/hotwired/stimulus/pull/546
[toggle]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement/toggle_event
[turbo-events]: https://turbo.hotwired.dev/reference/events

Preview of changes to the `Actions` documentation page:
===

You can register your own action options with the `Application.registerActionOption` method.

 For example, consider that a `<details>` element will dispatch a [toggle][]
 event whenever it's toggled. A custom `:open` action option would help
 to route events whenever the element is toggled _open_:

 ```javascript

 application.registerActionOption("open", ({ event }) => {
   if (event.type == "toggle") {
     return event.target.open == true
   } else {
     return true
   }
 })
 ```

 Similarly, a custom `:!open` action option could route events whenever the
 element is toggled _closed_. Declaring the action descriptor option with a `!`
 prefix will yield a `value` argument set to `false` in the callback:

 ```javascript

 application.registerActionOption("open", ({ event, value }) => {
   if (event.type == "toggle") {
     return event.target.open == value
   } else {
     return true
   }
 })
 ```

 In order to prevent the event from being routed to the controller action, the
 `registerActionOption` callback function must return `false`. Otherwise, to
 route the event to the controller action, return `true`.

 The callback accepts a single object argument with the following keys:

 Name    | Description
 --------|------------
 name    | String: The option's name (`"open"` in the example above)
 value   | Boolean: The value of the option (`:open` would yield `true`, `:!open` would yield `false`)
 event   | [Event][]: The event instance
 element | [Element]: The element where the action descriptor is declared

 [toggle]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement/toggle_event
 [Event]: https://developer.mozilla.org/en-US/docs/web/api/event
 [Element]: https://developer.mozilla.org/en-US/docs/Web/API/element

### DIFF
--- a/docs/reference/actions.md
+++ b/docs/reference/actions.md
@@ -109,6 +109,61 @@ Custom action option | Description
 `:prevent`           | calls `.preventDefault()` on the event before invoking the method
 `:self`              | only invokes the method if the event was fired by the element itself
 
+You can register your own action options with the `Application.registerActionOption` method.
+
+For example, consider that a `<details>` element will dispatch a [toggle][]
+event whenever it's toggled. A custom `:open` action option would help
+to route events whenever the element is toggled _open_:
+
+```javascript
+import { Application } from "@hotwired/stimulus"
+
+const application = Application.start()
+
+application.registerActionOption("open", ({ event }) => {
+  if (event.type == "toggle") {
+    return event.target.open == true
+  } else {
+    return true
+  }
+})
+```
+
+Similarly, a custom `:!open` action option could route events whenever the
+element is toggled _closed_. Declaring the action descriptor option with a `!`
+prefix will yield a `value` argument set to `false` in the callback:
+
+```javascript
+import { Application } from "@hotwired/stimulus"
+
+const application = Application.start()
+
+application.registerActionOption("open", ({ event, value }) => {
+  if (event.type == "toggle") {
+    return event.target.open == value
+  } else {
+    return true
+  }
+})
+```
+
+In order to prevent the event from being routed to the controller action, the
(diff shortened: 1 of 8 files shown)


dhh:
Love it! And I actually love the two examples you're giving enough that I think they're worth adding as defaults too.

seb-jean:
Thank you for this PR.

How could we do with a [dialog](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement) element to add a custom `:open` action?

PR #571, opened 2022-08-09
Make Action Parameters attributes case-insensitive

### LINKED ISSUE #561: Action parameters do not work if the controller identifier has any capital letter in it
Hi Stimulus team, and thanks for your amazing work.

Sorry for not having replied on time, but referring to my previous issue: https://github.com/hotwired/stimulus/issues/558 I think I did not explain clearly my problem, so I will try to do it now.

Problem: when the controller identifier has any capital letter in it, params are not passed to the triggered action.

This is a code snippet that reproduces the issue:
```
<html>
<head>
  <script type="module">
    import { Application, Controller } from "https://cdn.skypack.dev/@hotwired/stimulus@3.0.1";

    class CamelCaseController extends Controller {
        action(event) {
            alert(event.params.foo); // Opens alert with `undefined`
        }
    };

    application.register("CamelCase", CamelCaseController);
  </script>
</head>
<body>
 <div data-controller="CamelCase">
   <button data-action="CamelCase#action"
        data-CamelCase-foo-param="bar">
     Open alert
   </button>
 </div>
</body>
</html>
```

The result is that `undefined` is printed, even though the parameter `foo` should have value `bar`

If you change the controller name to `camelcase`, thus removing capital letters both in html and js, things will work as documented.

I believe this is an issue because other features of the framework (i.e. targets and values) work even when the controller identifier is not all lowercase.

If you confirm this is something to be fixed, I can open a PR for this.

Thanks!

marcoroth:
Even though this is not the "recommended way" to specify the data 
attributes for the params it's to match the existing behaviour of 
Actions, Controllers, Targets and Values.

Resolves #561

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -31,7 +31,7 @@ export class Action {
 
   get params() {
     const params:{ [key: string]: any } = {}
-    const pattern = new RegExp(`^data-${this.identifier}-(.+)-param$`)
+    const pattern = new RegExp(`^data-${this.identifier}-(.+)-param$`, "i")
 
     for (const { name, value } of Array.from(this.element.attributes)) {
       const match = name.match(pattern)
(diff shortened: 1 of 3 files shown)


adrienpoly (approved):
👌

PR #575, opened 2022-08-19
Change != to !==

searls:
Use the [strict inequality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality) operator instead of the looser one that will convert operands to compare them. 

Just saw this and thought I'd share because strict equality/inequality is generally preferred and recommended.

See https://github.com/hotwired/stimulus-site/pull/30


marcoroth (approved):
Thank you @searls!

PR #577, opened 2022-08-30
Add: clarifying example

benkoshy:
### Why?

* examples are helpful.
* I often get confused/forget. 
* Perhaps others may benefit?

### DIFF
--- a/docs/reference/targets.md
+++ b/docs/reference/targets.md
@@ -121,4 +121,15 @@ with a matching name, the corresponding callback _will not_ be invoked again.
 
 ## Naming Conventions
 
-Always use camelCase to specify target names, since they map directly to properties on your controller.
+Always use camelCase to specify target names, since they map directly to properties on your controller:
+
+```html
+<span data-search-target="camelCase" /> 
+<span data-search-target="do-not-do-this" />
+```
+
+```js
+export default class extends Controller {
+  static targets = [ "camelCase" ]  
+}
+```


docs/reference/targets.md
```diff
@@ -121,4 +121,15 @@ with a matching name, the corresponding callback _will not_ be invoked again.
 
 ## Naming Conventions
 
-Always use camelCase to specify target names, since they map directly to properties on your controller.
+Always use camelCase to specify target names, since they map directly to properties on your controller:
+
+```erb
```
lb-, line 126: maybe make this `html` as not everyone who uses stimulus uses ruby.

Also use HTML syntax

docs/reference/targets.md
```diff
@@ -121,4 +121,15 @@ with a matching name, the corresponding callback _will not_ be invoked again.
...
+<span data-search-target="camelCase" span/> 
+<span data-search-target="do-not-do-this" span/>
```
marcoroth, line 128:
```suggestion
<span data-search-target="camelCase" /> 
<span data-search-target="do-not-do-this" />
```

docs/reference/targets.md
```diff
@@ -121,4 +121,15 @@ with a matching name, the corresponding callback _will not_ be invoked again.
...
+```
+
+```js
+export default class extends Controller {
+  static targets = ["camelCase" ]
```
marcoroth, line 133:
```suggestion
  static targets = [ "camelCase" ]  
```

docs/reference/targets.md
benkoshy: Oops. Nice catch.

docs/reference/targets.md
benkoshy: will force push. nice catch.

marcoroth (approved):
Thank you @benkoshy!

Issue #574, opened 2022-08-17
Feature request - Controller registration (load) callback 


lb-:
## Summary

When a controller is registered (aka loaded), it would be helpful to provide the ability for the controller (class) to be able to provide a callback to know it has been registered and with what identifier.

## Details

- There are cases where the controller may want to cause some side effects (mutate DOM, load some expensive JS, check something in the application) once it has been registered. Without having to wait for the first 'connect' instance method bring called. 
- The `shoudLoad` static get method can kind of serve this purpose but this has no access to the application instance, nor the identifier string that the controller has been registered with.
- This becomes very useful when providing controllers as a standalone package - the controller will not be able to know the application (schema, logger, Dom element attached to) nor the identifier name used. Nor will the controller really be able to know that an instance of it will ever be created, but may want to do some setup or logic once registered.
- Long term, an unregister static method could also be supported. 
- I am happy to work on implementation of this feature if it seems useful to others. 

## Example usage - adding support for the controller to bootstrap its own usage for classes triggering 

- Provide a way to trigger some behaviour once the controller has been registered with an application. 
- In the case below we want to find any legacy usage of classes that do not yet have the new data controller code and attach a controller. 
- We also need to know the application's controller attribute name so that this code can work for any custom schema 

```
class SpinnerButton extends Controller {
  static register(identifier, application) {
    // Add check for Dom state ready ...
   document.querySelctor('.button.spinner').forEach(elem => {
    elem.setAttribute(application.schema.controllerAttribute, identifier);
   });
  }
}
```

## Example usage - side effects before DOM is ready

- A controller may have some behaviour that is associated with it but must be set before the Dom is ready for a better user experience
- This is an example of a dark mode toggle controller where it will run some code as soon as possible once registered. 
- it also relies on the identifier to prepare a usage specific session storage key

```
class DarkModeToggleButton extends Controller {
  static register(identifier, application) {
     // do not check for DOM ready - as we do not want a flash of content
     const savedThemeMode = localStorage.getItem(`${identifier}-theme');
    const prefersDarkMode = window.matchMedia(
      "(prefers-color-scheme:dark)"
    ).matches;
    console.log('apply saved or preferred dark mode');
  }

  constructor() {
    this.storageKey = `${this.identifier}-theme';
  }

  toggle() {
     document.body.class = ''; 
     // Toggle dark mode' class & update session storage. 
  }
}
```

## Alternative approaches

[...]

PR #579, opened 2022-08-30
Add ability to set `afterLoad` static methods on Controllers

lb-:
- When a controller is registered, the `afterLoad` static method, if present, will be called
- It gets passed the application instance and the identifier that was used to register it
- implements https://github.com/hotwired/stimulus/issues/574

Note: Hopefully I have added things correctly, I could not get the examples working so have only added unit tests.

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -163,6 +163,38 @@ class UnloadableController extends ApplicationController {
 application.register("unloadable", UnloadableController)
 ```
 
+### Trigger Behaviour When A Controller Is Registered
+
+If you want to trigger some behaviour once a controller has been registered you can add a static `afterLoad` method:
+
+```js
+class SpinnerButton extends Controller {
+  static afterLoad(identifier, application) {
+    // use the application instance to read the configured 'data-controller' attribute
+    const { controllerAttribute } = application.schema
+
+    // update any legacy buttons with the controller's registered identifier
+    const updateLegacySpinners = () => {
+      document.querySelector(".legacy-spinner-button").forEach((element) => {
+        element.setAttribute(controllerAttribute, identifier)
+      })
+    }
+
+    // called as soon as registered so DOM many not have loaded yet
+    if (document.readyState == "loading") {
+      document.addEventListener("DOMContentLoaded", updateLegacySpinners)
+    } else {
+      updateLegacySpinners()
+    }
+  }
+}
+
+// This controller will update any legacy spinner buttons to use the controller
+application.register("spinner-button", SpinnerButton)
+```
+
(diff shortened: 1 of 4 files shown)


rik:
`afterLoad` makes me think this will be called after the [`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) event. `afterRegister` seems less confusing to me.

lb-:
> `afterLoad` makes me think this will be called after the [`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) event. `afterRegister` seems less confusing to me.

Yeah - see https://github.com/hotwired/stimulus/issues/574#issuecomment-1221374478

Basically, the current static method `shouldLoad` is already in place.

While `afterRegister` better aligns with the overall nomenclature of `register` but then there will be two static methods; `shouldLoad` and `afterRegister` that will be inconsistent.

A good thing to note is that you can also load a suite of controllers via `application.load([...someControllersWithIdentifiers])` - I think that is why the naming is inconsistent.

`application.register` only takes one controller (and calls `application.load` under the hood), `application.load` takes one or more controllers.

If we opt for `afteRegister`, I think we should also move to `shouldRegister` with handling for a legacy fallback of `shouldLoad` (although, this is a bit tricky due to the boolean check and the fact it is a static get method, so checking for its presence needs to be done carefully).

marcoroth:
> `afterLoad` makes me think this will be called after the [`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) event. `afterRegister` seems less confusing to me.

What if we call it `afterApplicationLoad()`?

lb-:
I think afterApplicationLoad could be confusing in a different way. It may be inferred to be after the application has loaded instead of after this controller is loaded by the application. 

What if I Update to afterRegister or afterRegistered ?

Then can explore a rename if shouldLoad to should register as a seperate issue.

marcoroth:
Looking at this function I can totally understand why it's called `shouldLoad`, it's just super confusing because you usually don't  call `load()` in your application
https://github.com/hotwired/stimulus/blob/da425c43c8293af860105fdd1647832fa63416c9/src/core/application.ts#L56-L65

lb-:
The documentation does reference `application.load` when used with the webpack helpers. 

https://stimulus.hotwired.dev/handbook/installing#using-webpack-helpers

lb-:
Note to self - rebase on master now that eslint/prettier is adopted.

dhh:
Went with shouldLoad because it's called as part of load. Even though the user invocation happens through register. I think @lb- is right to keep this consistent. Also don't think there's enough value in a rename to proceed with a depreciation cycle here to be honest.

PR #584, opened 2022-09-12
Retain backtrace for TypeErrors in value change callback

elliotcm:
This code was originally added to provide better error messaging for TypeErrors when assigning values.

Unfortunately throwing a new error object causes us to lose the context (including stacktrace) of the original error, making it harder to locate the source.

Changing the message and rethrowing the original error retain that information.

## Before

Screenshot 2022-09-11 at 22 01 34

## After

Screenshot 2022-09-11 at 22 11 27

Thanks to @marcoroth for help discussing this issue.


marcoroth (approved):
Thank you, @elliotcm!

PR #590, opened 2022-10-13
adding Default Values

ThomasLandauer:


dhh:
Again, there just isn't enough here to warrant this.

ThomasLandauer:
I was looking for default values, and didn't expect them to be hidden under "Getters"

PR #591, opened 2022-10-16
Moving Object upwards

ThomasLandauer:
Topical ordering is more useful than alphabetical.


dhh:
Don't see the point on such a small table tbh.

ThomasLandauer:
It improves scanability if both "JSON" entries are next to each other.

Issue #489, opened 2021-11-12
Memory leak


vastray:
Removed elements are not deleted from memory.

![Снимок экрана от 2021-11-12 13-47-05](https://user-images.githubusercontent.com/51928487/141451228-794e626c-b00e-4512-b4f1-04821bc5e729.png)
^^^ Nodes and Listeners are not cleared

We create the "List" controller, in it we create elements for the "Item" controller. When the button is clicked, the item is cleared and a list is created. When you click on the "Reload" button, a memory leak occurs. The garbage collector doesn't help.

Example:

index.html
```html
<div data-controller="list">
    <button data-action="list#reload">Reload</button>
    <div data-list-target="listContainer"></div>
</div>
```

script.js
```javascript
const data = ['name1', 'name2', 'name3', 'name4', 'name5', 'name6', 'name7', 'name8']

Stimulus.register("list", class extends Controller {

    static targets = ['listContainer']

    connect() {
        this.load()
    }

    load() {
        data.forEach((name) => {
            const div = document.createElement('div')
            div.setAttribute('data-controller', 'item')
            div.setAttribute('data-action', 'click->item#log')
            div.innerText = name
            this.listContainerTarget.appendChild(div)
        })
    }

    reload() {
        this.listContainerTarget.innerHTML = ''
        this.load()
    }
})

Stimulus.register("item", class extends Controller {

    log() {
        console.log('click')
    }

})
```

tleish:
I'm pretty sure this has nothing to do with stimulus.  In this instance, stimulus is the event handling for executing the javascript, it the standard javascript that creating and clearing the nodes.

It's been discussed a lot in other javascript forums: 
- [javascript - Empty-ing innerHTML increases number of DOM nodes - Stack Overflow](https://stackoverflow.com/questions/22486030/empty-ing-innerhtml-increases-number-of-dom-nodes)
- [javascript - How to remove DOM elements without memory leaks? - Stack Overflow](https://stackoverflow.com/questions/3785258/how-to-remove-dom-elements-without-memory-leaks)
- [Deleting HTML Elements in JavaScript to avoid memory leaks - Stack Overflow](https://stackoverflow.com/questions/49147182/deleting-html-elements-in-javascript-to-avoid-memory-leaks)

If it's not an issue, you can ignore it.  Otherwise for you can read through this post for possible alternatives to innerHTML:
[...]

Issue #547, opened 2022-06-02
Potential Memory Leak in Dispatcher


redrockcity:
I'm attempting to use Stimulus to implement infinite scroll for a list of results. As the page is scrolled, new pages of results are fetched and their elements added to the DOM. Some of these elements have stimulus controllers, actions, or are targets. As these elements are added to the DOM, Stimulus is correctly attaching controllers and event handlers to them as expected.

As an optimization, as the page is scrolled further, results that have moved off screen are removed from the DOM. Again, Stimulus is correctly detaching controllers and tearing down event handlers as elements are removed.

If the page is scrolled back up to view results that were previously removed from the DOM, they are refetched and reinserted into the DOM. If the page is scrolled up and down repeatedly, result elements will be removed, refetched, and reinserted as needed. The effect is that only the results the user is currently viewing (with a small buffer) are present in the the DOM at any time.

While running some tests of scrolling the page up and down, I noticed the heap size of the page continually growing. I did a heap snapshot and noticed _a large amount of detached HTMLElements (representing the removed results) were still on the heap_. Although they were not in the DOM, _they were still referenced by the eventListenerMaps in the Stimulus dispatcher_.

![image](https://user-images.githubusercontent.com/25106557/171695652-10459eb5-c3fd-42ac-8831-b15bb6524205.png)

Over time, eventListenerMaps is continuing to grow with detached eventTargets even after they have been disconnected from a controller and removed from the DOM. My infinite scroll page certainly exacerbates this effect, however I would think it could still be an issue with long running SPAs or Turbo apps using Stimulus.

I've tried removing the result elements using parentElement.innerHtml = "", element.remove(), parentElement.removeChildren() but am seeing the same effect with eventListenerMaps. Is there a different way of removing elements such that Stimulus will not maintain a reference to them and they can be garbage collected?

Thanks!

redrockcity:
> Did you happen to read issue #489?

I did read that issue and attempted to implement the same fix mentioned at the bottom. I looped through all the elements and removed any data-* attributes prior to removing them from the DOM—thinking Stimulus might clean up the event listener mappings to said elements. Unless I did something wrong, it didn't seem to have any effect. I still got an accumulation of detached element references in the eventListenerMaps over prolonged use of the page.

intrip:
@redrockcity https://github.com/hotwired/stimulus/pull/592 might have fixed this, can you still reproduce using the latest StimulusJS version?

marcoroth:
[...]

PR #592, opened 2022-10-20
Clear dangling EventListeners and Detached Nodes when a controller is removed from the DOM

intrip:
`eventListenerMaps` maps HTML elements handled by Stimulus to `eventListenerMap`, a map of Stimulus EventListeners in this way: `{ HTMLelement: { eventName: EventListener } }`. When a controller HTML is removed from the DOM the dangling EventListeners aren't removed, moreover, the removed HTML elements are still referenced via the `eventListenerMaps` keys. This leads to having multiple detached HTMLelements and eventListeners which can't be GCed hence a memory leak. The leak is fixed by removing the dangling eventListeners and clearing unused `eventListenerMaps` keys. When the HTMLelement is attached again to the DOM, the `eventListenerMaps` and the related `eventListeners` are automatically re-created by Stimulus via the MutationObservers, no data is lost by doing this.

Below is an example to reproduce the issue:

```html
<html>
  <head>
    <script type="module">
        doClick() { process(500) }
      })
    </script>
    <script>
      function process(count) {
        let i = 0
        let handler = setInterval(function(){
          if (++i > count) {
            clearInterval(handler)
          } else {
            document.body.replaceWith(document.body.cloneNode(true))
          }
        }, 1)
      }
    </script>
</head>
<body>
    To reproduce:
    <ul>
      <li>Check heap snapshot</li>
    </ul>
  </div>
</body>
</html>
```

heap snapshot before:

Screenshot 2022-10-19 at 17 10 35

heap snapshot after:

Screenshot 2022-10-19 at 17 11 10

**Related** https://github.com/hotwired/stimulus/issues/489 https://github.com/hotwired/stimulus/issues/547

### DIFF
--- a/src/core/binding_observer.ts
+++ b/src/core/binding_observer.ts
@@ -7,7 +7,7 @@ import { Token, ValueListObserver, ValueListObserverDelegate } from "../mutation
 
 export interface BindingObserverDelegate extends ErrorHandler {
   bindingConnected(binding: Binding): void
-  bindingDisconnected(binding: Binding): void
+  bindingDisconnected(binding: Binding, clearEventListeners?: boolean): void
 }
 
 export class BindingObserver implements ValueListObserverDelegate<Action> {
@@ -72,7 +72,7 @@ export class BindingObserver implements ValueListObserverDelegate<Action> {
   }
 
   private disconnectAllActions() {
-    this.bindings.forEach((binding) => this.delegate.bindingDisconnected(binding))
+    this.bindings.forEach((binding) => this.delegate.bindingDisconnected(binding, true))
     this.bindingsByAction.clear()
   }
 
--- a/src/core/dispatcher.ts
+++ b/src/core/dispatcher.ts
@@ -41,14 +41,32 @@ export class Dispatcher implements BindingObserverDelegate {
     this.fetchEventListenerForBinding(binding).bindingConnected(binding)
   }
 
-  bindingDisconnected(binding: Binding) {
+  bindingDisconnected(binding: Binding, clearEventListeners: boolean = false) {
     this.fetchEventListenerForBinding(binding).bindingDisconnected(binding)
+    if (clearEventListeners) this.clearEventListenersForBinding(binding)
   }
 
   // Error handling
 
   handleError(error: Error, message: string, detail: object = {}) {
     this.application.handleError(error, `Error ${message}`, detail)
+  }
+
+  private clearEventListenersForBinding(binding: Binding) {
+    const eventListener = this.fetchEventListenerForBinding(binding)
+    if (!eventListener.hasBindings()) {
+      eventListener.disconnect()
+      this.removeMappedEventListenerFor(binding)
+    }
+  }
+
+  private removeMappedEventListenerFor(binding: Binding) {
+    const { eventTarget, eventName, eventOptions } = binding
+    const eventListenerMap = this.fetchEventListenerMapForEventTarget(eventTarget)
+    const cacheKey = this.cacheKey(eventName, eventOptions)
+
+    eventListenerMap.delete(cacheKey)
+    if (eventListenerMap.size == 0) this.eventListenerMaps.delete(eventTarget)
   }
 
   private fetchEventListenerForBinding(binding: Binding): EventListener {
--- a/src/core/event_listener.ts
+++ b/src/core/event_listener.ts
@@ -43,6 +43,10 @@ export class EventListener implements EventListenerObject {
     }
   }
 
+  hasBindings() {
+    return this.unorderedBindings.size > 0
+  }
+
   get bindings(): Binding[] {
     return Array.from(this.unorderedBindings).sort((left, right) => {
       const leftIndex = left.index,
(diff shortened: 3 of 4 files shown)


rik:
Could this be solved by only changing `Dispatcher.eventListenerMaps` to a `WeakMap`?

https://github.com/hotwired/stimulus/blob/1dc541567568ac1f835dcf6cabda6952a84ce6b2/src/core/dispatcher.ts#L13

intrip:
> Could this be solved by only changing `Dispatcher.eventListenerMaps` to a `WeakMap`?
> 
> https://github.com/hotwired/stimulus/blob/96869438506e23ab9ccda5df4b2a96fa4ce0f55c/src/core/dispatcher.ts#L13

I thought about this too but, unfortunately, I don't think it would work: `EventListener` objects (referenced via [EventListenerMaps](https://github.com/hotwired/stimulus/blob/6bb80e4e46b7bc83def80390efee0eb701562cf9/src/core/dispatcher.ts#L8)  have a specific lifecycle so they need to be correctly [disconnected](https://github.com/hotwired/stimulus/blob/6bb80e4e46b7bc83def80390efee0eb701562cf9/src/core/event_listener.ts#L31) before being removed, by using a WeakMap we don't have control on that. For example, If an `EventListener` won't be disconnected properly we'll have a case when the Dispatcher will [re-create and connect a new one](https://github.com/hotwired/stimulus/blob/44943a363522806d8be43d98f85e7a151865be59/src/core/dispatcher.ts#L75-L78) and we'll end up in a state with duplicate EventListeners connected.

src/core/binding_observer.ts
```diff
@@ -7,7 +7,7 @@ import { Token, ValueListObserver, ValueListObserverDelegate } from "../mutation
 
 export interface BindingObserverDelegate extends ErrorHandler {
   bindingConnected(binding: Binding): void
-  bindingDisconnected(binding: Binding): void
+  bindingDisconnected(binding: Binding, clearEventListeners?: boolean): void
```
jorgemanrubia, line 10: Is there a scenario where we want to invoke this `bindingDisconnected` method without clearing the listeners?

src/core/dispatcher.ts
```diff
@@ -51,6 +52,19 @@ export class Dispatcher implements BindingObserverDelegate {
...
+    if (eventListener.bindings.length == 0) {
+      const { eventTarget, eventName, eventOptions } = binding
+      const eventListenerMap = this.fetchEventListenerMapForEventTarget(eventTarget)
+      const cacheKey = this.cacheKey(eventName, eventOptions)
+      eventListener.disconnect()
+      eventListenerMap.delete(cacheKey)
+
+      if (eventListenerMap.size == 0) this.eventListenerMaps.delete(eventTarget)
+    }
+  }
```
jorgemanrubia, line 66: Maybe could extract a couple of methods to clarify logic. I also added a method `EventListener.hasBindings`:

```ts
private clearEventListenersForBinding(binding: Binding) {
  const eventListener = this.fetchEventListenerForBinding(binding)
  if (!eventListener.hasBindings()) {
    this.removeMappedEventListenerFor(binding)
    eventListener.disconnect()
  }
}

private removeMappedEventListenerFor(binding: Binding) {
  const { eventTarget, eventName, eventOptions } = binding
  const eventListenerMap = this.fetchEventListenerMapForEventTarget(eventTarget)
  const cacheKey = this.cacheKey(eventName, eventOptions)

  eventListenerMap.delete(cacheKey)
  if (eventListenerMap.size == 0) this.eventListenerMaps.delete(eventTarget)
}
```

src/core/dispatcher.ts
```diff
@@ -51,6 +52,19 @@ export class Dispatcher implements BindingObserverDelegate {
     this.application.handleError(error, `Error ${message}`, detail)
   }
 
+  private clearEventListenersForBinding(binding: Binding) {
+    const eventListener = this.fetchEventListenerForBinding(binding)
+    if (eventListener.bindings.length == 0) {
```
jorgemanrubia, line 57: Why do we need this check `eventListener.bindings.length == 0`? In which scenario can this be invoked where this doesn't validate? I don't see anything wrong I just want to understand 😅.

jorgemanrubia (commented):
Amazing work tracking this one down @intrip 🙌!

src/core/dispatcher.ts
intrip: Please ask! 
It's a rough edge but It's possible to have multiple bindings for the same `EventListener`. Let's say we have this scenario:

```
  <button data-action="click->boosts#doSomething click->boosts#doSomething">Hey!</button>
</div>
```

In this case, if change the "data-action" we don't want to disconnect the `EventListener` unless we remove both "click->boosts#doSomething" actions.
Right now we call this method only if the whole controller is removed so the safeguard is not a necessity, it was a necessity in the first version I made which didn't use the "clearEventListeners" flag.

intrip:
> Great, this would be able to potentially fix #489 and #547?

Yes, I think those two issues are related.

PR #603, opened 2022-11-17
Ensure that the Application.start static method uses overridden class

### LINKED ISSUE #586: Ability to extend the Stimulus Application class
## Summary 

Class inheritance provides a way to achieve custom behaviour and it would be great to also allow this at the `Application` level. 

Currently, it is not easy to extend and use the `Application` class without rewriting its `static start` method. 

## Details 

In the Application code, the `start` static method references the `Application` class directly instead of using `this`. 

https://github.com/hotwired/stimulus/blob/f690a6a2e856a35ebdb78246285b734da541338e/src/core/application.ts#L19-L23

This means that even if you extend the `Application` with say a custom method override of `load`, this custom method will never get used. 

## Proposal 

Using `this` instead of `Application` should solve the issue. 

```Typescript
  static start(element?: Element, schema?: Schema): Application {
    const application = new this(element, schema)
    application.start()
    return application
  }
```

## Use case

A trivial use case to add extra behaviour to the application stop call. 

```javascript
class MyApplication extends Application {
  stop() {
    super();
    console.log('application stopped');
  }
}

const application= MyApplication.start();
application.stop();
```

lb-:
- When using `const application = MyApplication.start()` - it now will use the actual `MyApplication` class instead of the non-extended `Application` class
- Fixes #586

## Notes

I opted for no documentation here - I think this behaviour would be expected and also aligns with any manual `start` usage with instantiation of the class yourself.

For example, the following two ways of starting the application will now be the same (with this PR merged).

### A - Starting an application as per documentation

```javascript
class MyApplication extends Application {}

const element = document.getElementById('main');
const application = MyApplication.start(element); // without this PR - the `application` will be an instance of the Stimulus `Application` and not `MyApplication`
```

### B - Setting up the instance manually & running `start`

```javascript
class MyApplication extends Application {}

const element = document.getElementById('main');
const application = new MyApplication(element);
application.start();
```

## Test format

Note that I had to write my own test case class as `src/tests/cases/application_test_case.ts` does not use the documented `Application.start()` approach (due to other tests relying on a separate setup step for the application before starting).

This is similar to the way that the test file `src/tests/modules/core/error_handler_tests.ts` also sets up its own test case but without the double-registration of a Stimulus application. Happy to adjust if this is not the way we want to do this.

### DIFF
--- a/src/core/application.ts
+++ b/src/core/application.ts
@@ -17,7 +17,7 @@ export class Application implements ErrorHandler {
   debug = false
 
   static start(element?: Element, schema?: Schema): Application {
-    const application = new Application(element, schema)
+    const application = new this(element, schema)
     application.start()
     return application
   }
--- a/src/tests/modules/core/extending_application_tests.ts
+++ b/src/tests/modules/core/extending_application_tests.ts
@@ -1,0 +1,51 @@
+import { Application } from "../../../core/application"
+import { DOMTestCase } from "../../cases/dom_test_case"
+import { ActionDescriptorFilter } from "src/core/action_descriptor"
+
+const mockCallback: { (label: string): void; lastCall: string | null } = (label: string) => {
+  mockCallback.lastCall = label
+}
+mockCallback.lastCall = null
+
+class TestApplicationWithCustomBehavior extends Application {
+  registerActionOption(name: string, filter: ActionDescriptorFilter): void {
+    mockCallback(`registerActionOption:${name}`)
+    super.registerActionOption(name, filter)
+  }
+}
+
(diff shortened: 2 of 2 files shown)


lb-:
I do not think the error is due to this PR's changes (tests ran ok locally).

>  17 11 2022 21:36:45.021:ERROR [SaucelabsLauncher]: Error: 17 Nov 21:36:45 - couldn't connect to https://saucelabs.com/rest/v1/stimulus/tunnels: Post https://saucelabs.com/rest/v1/stimulus/tunnels: net/http: timeout awaiting response headers

lb-:
Thanks @dhh

PR #604, opened 2022-11-18
Documentation for Outlets API

marcoroth:
This Pull request adds documentation for the Outlets API which was introduced in #576.

It also updates the `order` for the other sections in the reference.

Feedback very welcome, especially on the used examples.

### DIFF
--- a/docs/reference/outlets.md
+++ b/docs/reference/outlets.md
@@ -1,0 +1,179 @@
+---
+permalink: /reference/outlets.html
+order: 04
+---
+
+# Outlets
+
+_Outlets_ let you reference Stimulus _controller instances_ and their _controller element_ from within another Stimulus Controller by using CSS selectors.
+
+The use of outlets helps with cross-controller communication and coordination as an alternative to dispatching custom events on controller elements.
+
+They are conceptually similar to [Stimulus Targets](https://stimulus.hotwired.dev/reference/targets) but with the difference that they reference a Stimulus controller instance plus its associated controller element.
+
+<meta data-controller="callout" data-callout-text-value='data-search-result-outlet=".result"'>
+<meta data-controller="callout" data-callout-text-value='class="result"'>
+
+
+```html
+<div data-controller="search" data-search-result-outlet=".result">
+  ...
+</div>
+
+...
+
+<div id="results">
+  <div class="result" data-controller="result">...</div>
+  <div class="result" data-controller="result">...</div>
+  ...
+</div>
+```
+
+While a **target** is a specifically marked element **within the scope** of its own controller element, an **outlet** can be located **anywhere on the page** and doesn't necessarily have to be within the controller scope.
+
+## Attributes and Names
+
+The `data-search-result-outlet` attribute is called an _outlet attribute_, and its value is a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) which you can use to refer to other controller elements which should be available as outlets on the _host controller_.
+
+```html
+data-[identifier]-[outlet]-outlet="[selector]"
+```
+
+<meta data-controller="callout" data-callout-text-value='data-search-result-outlet=".result"'>
+
+
+```html
+<div data-controller="search" data-search-result-outlet=".result"></div>
+```
+
+## Definitions
+
+Define controller identifiers in your controller class using the `static outlets` array. This array declares which other controller identifiers can be used as outlets on this controller:
+
+<meta data-controller="callout" data-callout-text-value='static outlets'>
+<meta data-controller="callout" data-callout-text-value='"result"'>
+
+
+```js
+// search_controller.js
+
+export default class extends Controller {
+  static outlets = [ "result" ]
+
+  connect () {
+    this.resultOutlets.forEach(result => ...)
+  }
+}
+```
+
+## Properties
+
+For each outlet defined in the `static outlets` array, Stimulus adds five properties to your controller, where `[name]` corresponds to the outlet's controller identifier:
+
+| Kind | Property name | Return Type | Effect
+| ---- | ------------- | ----------- | -----------
+| Existential | `has[Name]Outlet` | `Boolean` | Tests for presence of a `[name]` outlet
+| Singular | `[name]Outlet` | `Controller` | Returns the `Controller` instance of the first `[name]` outlet or throws an exception if none is present
+| Plural | `[name]Outlets` | `Array<Controller>` | Returns the `Controller` instances of all `[name]` outlets
+| Singular | `[name]OutletElement` | `Element` | Returns the Controller `Element` of the first `[name]` outlet or throws an exception if none is present
+| Plural | `[name]OutletElements` | `Array<Element>` | Returns the Controller `Element`'s of all `[name]` outlets
+
+## Accessing controllers and elements
+
+Since you get back a `Controller` instance from the `[name]Outlet` and `[name]Outlets` properties you are also able to access the Values, Classes, Targets and all of the other properties and functions that controller instance defines:
+
+```js
+this.resultOutlet.idValue
+this.resultOutlet.imageTarget
+this.resultOutlet.activeClasses
+```
+
+You are also able to invoke any function the outlet controller may define:
+
+```js
+// result_controller.js
+
+export default class extends Controller {
+  markAsSelected(event) {
+    // ...
+  }
+}
+
+// search_controller.js
+
+export default class extends Controller {
+  static outlets = [ "result" ]
+
+  selectAll(event) {
+    this.resultOutlets.forEach(result => result.markAsSelected(event))
+  }
+}
+```
+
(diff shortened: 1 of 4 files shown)


docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
+---
+permalink: /reference/outlets.html
+order: 04
+---
+
+# Outlets
+
+_Outlets_ let you reference Stimulus _controller instances_ and their _controller element_ from within another Stimulus Controller by using CSS selectors.
+
+The use of outlets helps with cross-controller communication and coordination as an alternative to dispatching custom events on controller elements.
```
lb-, line 10:
```suggestion
The use of Outlets helps with cross-controller communication and coordination as an alternative to dispatching custom events on controller elements.
```

It appears that Outlet is a proper noun throughout the rest of the page so maybe it should be here also

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+| Kind | Property name | Return Type | Effect
+| ---- | ------------- | ----------- | -----------
+| Existential | `has[Name]Outlet` | `Boolean` | Tests for presence of a `[name]` outlet
+| Singular | `[name]Outlet` | `Controller` | Returns the `Controller` instance of the first `[name]` outlet or throws an exception if none is present
+| Plural | `[name]Outlets` | `Array<Controller>` | Returns the `Controller` instances of all `[name]` outlets
+| Singular | `[name]OutletElement` | `Element` | Returns the Controller `Element` of the first `[name]` outlet or throws an exception if none is present
+| Plural | `[name]OutletElements` | `Array<Element>` | Returns the Controller `Element`'s of all `[name]` outlets
+
+## Accessing controllers and elements
```
lb-, line 81:
```suggestion
## Accessing Controllers and Elements
```

Most of the documentation appears to use APA title case (major words are title case).

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+Similarly with the Outlet Element, it allows you to call any function or property on [`Element`](https://developer.mozilla.org/en-US/docs/Web/API/Element):
+
+```js
+this.resultOutletElement.dataset.value
+this.resultOutletElement.getAttribute("id")
+this.resultOutletElements.map(result => result.hasAttribute("selected"))
+```
+
+## Outlet callbacks
```
lb-, line 121:
```suggestion
## Outlet Callbacks
```

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+Outlet callbacks are specially named functions called by Stimulus to let you respond to whenever an outlet is added or removed from the page.
+
+To observe outlet changes, define a function named `[name]OutletConnected()` or `[name]OutletDisconnected()`.
```
lb-, line 125:
```suggestion
To observe outlet changes, define a class method named `[name]OutletConnected()` or `[name]OutletDisconnected()`.
```

Not sure if needed but it may be clearer if we say that this is a class method?

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+    // ...
+  }
+
+  resultOutletDisconnected(outlet, element) {
+    // ...
+  }
+}
+```
+
+### Outlets are assumed to be present
```
lb-, line 143:
```suggestion
### Outlets are Assumed to be Present
```

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+When you access an outlet property in a controller, you assert that at least one corresponding outlet is present. If the declaration is missing and no matching outlet is found Stimulus will throw an exception:
```
lb-, line 145:
```suggestion
When you access an Outlet property in a Controller, you assert that at least one corresponding Outlet is present. If the declaration is missing and no matching outlet is found Stimulus will throw an exception:
```

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+```html
+Missing outlet element "result" for "search" controller
+```
+
+### Optional outlets
+
+If an outlet is optional or you want to assert that at least outlet is present, you must first check the presence of the outlet using the existential property:
```
lb-, line 153:
```suggestion
If an Outlet is optional or you want to assert that at least Outlet is present, you must first check the presence of the Outlet using the existential property:
```

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+```js
+if (this.hasResultOutlet) {
+  this.resultOutlet.safelyCallSomethingOnTheOutlet()
+}
+```
+
+### Referencing non-controller elements
```
lb-, line 161:
```suggestion
### Referencing Non-Controller Elements
```

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+
+Stimulus will throw an exception if you try to declare an element as an outlet which doesn't have the corresponding `data-controller` and identifier on it:
```
lb-, line 163: When will this error be thrown?

Once the controller is instantiated? Or when you attempt to access the outlet?

Might be good to be clear here - for some cases the data controller attribute could be added later maybe.

lb-, line 163:
```suggestion
Stimulus will throw an exception if you try to declare an element as an outlet which doesn't have a corresponding `data-controller` and identifier on it:
```

I'm not clear here - does the name of the Outlet's controller have to match the `static outlets = ['something']` name?

lb- (commented):
Some general comments below but what happens if you have an outlet that has two or more Controllers on it?

Maybe I missed something but does the name of the outlet have to match the name of the expected outlet controller?

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+  <div class="result" data-controller="result">...</div>
+  ...
+</div>
+```
+
+While a **target** is a specifically marked element **within the scope** of its own controller element, an **outlet** can be located **anywhere on the page** and doesn't necessarily have to be within the controller scope.
+
+## Attributes and Names
+
+The `data-search-result-outlet` attribute is called an _outlet attribute_, and its value is a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) which you can use to refer to other controller elements which should be available as outlets on the _host controller_.
```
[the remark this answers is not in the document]
marcoroth: Yeah agreed. I wasn't too happy with the term "host controller". It's not decided yet, that's just what I had in there as a placeholder. Happy to change it whatever makes to most sense. `source` seems better fitting to me

docs/reference/outlets.md
marcoroth: It's throwing the error when you attempt to access the outlet, similar to how targets handle this. I'll try to make this clearer, thanks!

marcoroth: It needs to match the name of the outlets, yes. The `outlets` array holds references to other controller identifiers.

docs/reference/outlets.md
marcoroth: Thanks for the review!

> Some general comments below but what happens if you have an outlet that has two or more Controllers on it?

An element with two controllers attached to it also has two different controllers instances. You'll get the controller instance for the controller identifier you referenced as the outlet. 

> Maybe I missed something but does the name of the outlet have to match the name of the expected outlet controller?

Yes, the outlet name has to match the controller identifier. This is what's used to get right controller instance for a given element. 
The reason why you are able to provide a CSS selector is that you can specify which element/instance you want to associate with the source controller, because usually there are more than one instance of the same controller on the page.

marcoroth:
Thanks for the review!

> Some general comments below but what happens if you have an outlet that has two or more Controllers on it?

An element with two controllers attached to it also has two different controllers instances. You'll get the controller instance for the controller identifier you referenced as the outlet. 

> Maybe I missed something but does the name of the outlet have to match the name of the expected outlet controller?

Yes, the outlet name has to match the controller identifier. This is what's used to get right controller instance for a given element. 
The reason why you are able to provide a CSS selector is that you can specify which element/instance you want to associate with the source controller, because usually there are more than one instance of the same controller on the page.

docs/reference/outlets.md
marcoroth: you are right, in the `targets` section they are also called "methods" 👍🏼

docs/reference/outlets.md
lb-: Thanks.

docs/reference/outlets.md
```diff
@@ -0,0 +1,179 @@
...
+<meta data-controller="callout" data-callout-text-value='data-search-result-outlet=".result"'>
+
+
+```html
+<div data-controller="search" data-search-result-outlet=".result"></div>
+```
+
+## Definitions
+
+Define controller identifiers in your controller class using the `static outlets` array. This array declares which other controller identifiers can be used as outlets on this controller:
```
lb-, line 51:
```suggestion
Define controller identifiers in your controller class using the `static outlets` array. This array declares which other Controller identifiers can be used as outlets on this controller:
```

I see we say identifiers here. Just another nitpicky fix for proper noun usage of Controller

PR #600, opened 2022-11-17
Upgrade `examples/` view files to comply with EJS 3.x

marcoroth:
Closes #588 

/cc @lb-

PR #606, opened 2022-11-22
convert examples from Turbolinks to Turbo

adrienpoly:
While working on another small PR I realized that the Stimulus example are still using Turbolinks. 
This PR migrates them to Turbo.


marcoroth (approved):
Thanks @adrienpoly, I also noticed it when I was working on #600, glad to have it updated now! 🥳

PR #609, opened 2022-11-28
Broken Sauce Labs link in README

kihaya:
I found Sauce Labs link in README returns 502, so I fix it to correct one.

スクリーンショット 2022-11-28 15 59 40


adrienpoly:
Thanks for spotting this

PR #610, opened 2022-11-28
Fix typo in error message

bradleypriest:
Was just reading through these latest changes, and spotted a minor typo from #442
Ping @NakajimaTakuya


NakajimaTakuya:
@bradleypriest  Thanks!

PR #616, opened 2022-12-02
Fixed error message in outlet_properties

drjayvee:
The error message hard-coded `data-controller` instead of referring to the schema's attribute name


marcoroth (approved):
Thank you @drjayvee!

PR #623, opened 2022-12-09
Expose extra exports

DEfusion:
We have a use case where we need some stimulus controllers to be able to interact with content within an iframe from outside the iframe. To achieve this we've created controllers that can deal with multiple scopes, this required us to extended a few classes that were not being exported so this change exposes those.


marcoroth:
Hey @DEfusion, thanks for opening this pull request!

I'm not really sure if we really want and should make these classes part of the public API

DEfusion:
@marcoroth Thanks for the quick response.

For a bit more detail we run a website builder with a WYSIWYG interface which we're currently upgrading in various ways and are using stimulus heavily to replace an SPA approach. In order to isolate the website from the editing interface we're containing it in an iframe, but rather than use iframe messages to connect the two I thought it would be really nice to allow stimulus controllers in the `top` window (where the editing UI lives) to interact seamlessly with content within the iframe.

I have achieved this by creating a base controller which has this ability by having a scope-like class and a context which are linked to the iframe. This allows the controller to be able to interact with targets, bindings etc. within both the `top` window and the iframe seamlessly. Here is the base controller for example:

```
import { Context, Controller, TargetSet } from "@hotwired/stimulus"
import { IframeContext } from "./iframe_context"
import { IframeScope } from "./iframe_scope"
import { TargetSets } from "./target_sets"

  readonly iframeScope: IframeScope
  readonly iframeTargets: TargetSet
  readonly targetSets: TargetSets
  readonly iframeContext: IframeContext

  constructor(context: Context) {
    super(context)
    this.iframeScope = new IframeScope(this.scope.schema, this.scope.identifier, this.scope.guide)
    this.iframeTargets = new TargetSet(this.iframeScope)
    this.targetSets = new TargetSets([this.scope.targets, this.iframeTargets])
    this.iframeContext = new IframeContext(this, this.iframeScope)
  }

  get targets() {
    return this.targetSets
  }

  get iframeElement() {
    return this.iframeScope.iframeElement;
  }

  get iframeDocument() {
    return this.iframeScope.iframeDocument;
  }

  get iframeDocumentElement() {
    return this.iframeScope.iframeDocumentElement;
  }

  get iframeWindow() {
    return this.iframeScope.iframeWindow;
  }
}
```

I mis-spoke when I said "extending" the classes that I've added as extra exports, in fact I'm just composing them using them in similar ways that context and scope use them but instantiating them within our iframe aware versions.

I know this might not be a useful change outside our exact use case, so if it isn't then I'm fine maintaining a fork for us with these extra exports but it would be less headache if I didn't need to.

marcoroth:
Hey @DEfusion 

Thanks for providing the details! Sadly I don't think it makes sense to publicly export these. However, what we could do is ship the individual JavaScript files in the package.

I published a dev-build to see if this would work for you. Since they are part of the private API they are subject to change. But this change at least would allow you to imported them out of the `dist/` at your own risk. Here's how you can test it:

```bash
yarn add @hotwired/stimulus@https://github.com/hotwired/dev-builds/archive/@hotwired/stimulus/9b10d6f.tar.gz
```


The imports would look something like:
```js
import { TargetSet } from "@hotwired/stimulus/dist/core/target_set"
import { Context } from "@hotwired/stimulus/dist/core/context"
```

And the types could be imported from:
```ts
import type { TargetSet } from "@hotwired/stimulus/dist/types/core/target_set"
import type { Context } from "@hotwired/stimulus/dist/types/core/context"
```

Let me know if that works for you.

DEfusion:
@marcoroth Thanks for that, including the source in the package is really helpful and works perfectly for our need. Thanks for being so accommodating.

DEfusion:
Thanks for that @marcoroth much appreciated.

PR #624, opened 2022-12-11
Outlets: Add observers for controller element attributes

seanpdoyle:
Extract `DomTestCase` helpers
===

In some cases, calls to `Element.setAttribute`,
`Element.removeAttribute`, `Node.appendChild`, or `Element.remove` must
be followed up with a call to `await this.nextFrame` so that Stimulus
has an opportunity to synchronize.

This commit introduces asynchronous helper method versions of those
calls that bake-in the subsequent call to `this.nextFrame`.

Outlets: Add observers for controller element attributes
===

With the current Outlet implementation, they're only ever connected or
disconnected when an element _matching_ the outlet selector connects or
disconnects. The selector _declared on the controller_ element can only
ever be set once: when it's connected. If that attribute ever changes
(for example, when it's set to a new value or removed entirely), the
outlets are not updated.

This commit adds test coverage to ensure the list of outlets and their
items is synchronized with changes on both sides: when matching elements
are connected and disconnected _as well as_ when the selector that
dictates whether or not they match is added, updated, or removed.

To do so, this commit extends the `OutletObserver` to also manage the
lifecycle of an `AttributeObserver` instance alongside its internally
managed `SelectorObserver`.

### DIFF
--- a/src/core/outlet_observer.ts
+++ b/src/core/outlet_observer.ts
@@ -1,64 +1,79 @@
 import { Multimap } from "../multimap"
+import { AttributeObserver, AttributeObserverDelegate } from "../mutation-observers"
 import { SelectorObserver, SelectorObserverDelegate } from "../mutation-observers"
 import { Context } from "./context"
 import { Controller } from "./controller"
 
 import { readInheritableStaticArrayValues } from "./inheritable_statics"
 
-type SelectorObserverDetails = { outletName: string }
+type OutletObserverDetails = { outletName: string }
 
 export interface OutletObserverDelegate {
   outletConnected(outlet: Controller, element: Element, outletName: string): void
   outletDisconnected(outlet: Controller, element: Element, outletName: string): void
 }
 
-export class OutletObserver implements SelectorObserverDelegate {
+export class OutletObserver implements AttributeObserverDelegate, SelectorObserverDelegate {
+  started: boolean
   readonly context: Context
   readonly delegate: OutletObserverDelegate
   readonly outletsByName: Multimap<string, Controller>
   readonly outletElementsByName: Multimap<string, Element>
   private selectorObserverMap: Map<string, SelectorObserver>
+  private attributeObserverMap: Map<string, AttributeObserver>
 
   constructor(context: Context, delegate: OutletObserverDelegate) {
+    this.started = false
     this.context = context
     this.delegate = delegate
     this.outletsByName = new Multimap()
     this.outletElementsByName = new Multimap()
     this.selectorObserverMap = new Map()
+    this.attributeObserverMap = new Map()
   }
 
   start() {
-    if (this.selectorObserverMap.size === 0) {
+    if (!this.started) {
       this.outletDefinitions.forEach((outletName) => {
-        const element = this.scope.element
-        const attributeName = this.outletAttributeFor(outletName)
-        const details: SelectorObserverDetails = { outletName }
-        const selectorObserver = new SelectorObserver(element, attributeName, document.body, this, details)
-
-        this.selectorObserverMap.set(outletName, selectorObserver)
+        this.setupSelectorObserverForOutlet(outletName)
+        this.setupAttributeObserverForOutlet(outletName)
       })
-
-      this.selectorObserverMap.forEach((observer) => observer.start())
-    }
-
-    this.dependentContexts.forEach((context) => context.refresh())
-  }
-
-  stop() {
-    if (this.selectorObserverMap.size > 0) {
-      this.disconnectAllOutlets()
-      this.selectorObserverMap.forEach((observer) => observer.stop())
-      this.selectorObserverMap.clear()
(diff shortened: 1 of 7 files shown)


seanpdoyle:
@marcoroth since you authored the original Outlets implementation, I'd really appreciate some code review on this.

IsraelDCastro (approved):

src/core/outlet_observer.ts
```diff
@@ -113,8 +113,8 @@ export class OutletObserver implements SelectorObserverDelegate {
 
   // Private
 
-  private selector(outletName: string) {
-    return this.scope.outlets.getSelectorForOutletName(outletName)
+  private outletAttributeFor(outletName: string) {
+    return this.context.schema.outletAttributeForScope(this.identifier, outletName)
```
marcoroth, line 117: I'm wondering if it makes sense that the `SelectorObserver` takes in an `attributeName` to extract the `selector` from instead of passing in the `selector` directly.

This makes the `SelectorObserver` dependent on an element/attribute to be present in order to be able to extract the selector from, so you couldn't just pass in a selector string anymore.

We currently don't have the need for this in the codebase, but I guess it would make sense to keep the responsibilities separated.

src/mutation-observers/selector_observer.ts
```diff
@@ -7,17 +8,25 @@ export interface SelectorObserverDelegate {
...
+  private readonly elementObserver: ElementObserver
+  private readonly delegate: SelectorObserverDelegate
+  private readonly matchesByElement: Multimap<string, Element>
+  private readonly details: object
+  private selector: string | null
+
+  constructor(
+    element: Element,
+    attributeName: string,
+    scope: Element,
```
marcoroth, line 22: This might be a bit nit-picky, but I feel like `scope` could be confused with the Stimulus [`Scope`](https://github.com/hotwired/stimulus/blob/47951cd011176f2c1a3dca0932d0c2bb79000892/src/core/scope.ts#L10) class here. 

Would it make sense to rename `element` to `controllerElement` and `scope` to `element`/`scopeElement`?

src/mutation-observers/selector_observer.ts
```diff
@@ -28,68 +37,119 @@ export class SelectorObserver implements ElementObserverDelegate {
...
   }
 
-  get element(): Element {
-    return this.elementObserver.element
+  elementUnmatchedAttribute() {
+    if (this.selector) {
+      const matchedElements = this.matchesByElement.getValuesForKey(this.selector)
+
+      for (const matchedElement of matchedElements) {
+        this.selectorUnmatched(matchedElement, this.selector)
```
marcoroth, line 70: Any reason why we are not destructing `selector` from `this` here as well?

marcoroth (commented):
Thanks @seanpdoyle for catching this oversight!

The outcome of this PR looks good to me. I left some comments and I'm curious to hear what you think.

src/mutation-observers/selector_observer.ts
```diff
@@ -19,14 +19,14 @@ export class SelectorObserver implements AttributeObserverDelegate, ElementObser
   constructor(
     element: Element,
     attributeName: string,
-    scope: Element,
+    controllerElement: Element,
```
marcoroth, line 22: I think it's switched up now. `controllerElement` is now `document.body` and `element` the actual controller element which is used for observing the attribute

src/core/outlet_observer.ts
seanpdoyle: Are you suggesting that the `OutletObserver` instance should be the one to listen for changes to the attribute, then somehow communicate with the `SelectorObserver` to refresh its set of elements?

marcoroth: I haven't actually thought this far. I was just confused that the `SelectorObserver` doesn't get an actual selector but needs to dig it out from somewhere, which technically shouldn't be it's responsibility, since it _could_ be used otherwise.

But now that you mentioned it, I guess it should be the responsibility of the `SelectorObserver `. Maybe we can extract the common logic so that the `OutletObserver` doesn't need to communicate with the `SelectorObserver ` to refresh the elements.

seanpdoyle: > Maybe we can extract the common logic so that the `OutletObserver` doesn't need to communicate with the `SelectorObserver`  to refresh the elements.

The changes proposed in https://github.com/hotwired/stimulus/pull/627 would benefit from a shared abstraction that watches an attribute with `document`-wide implications.

marcoroth: I implemented my proposal as discussed on Discord in https://github.com/hotwired/stimulus/pull/624/commits/4ff1250a04685ca9471f38cb37e3a06e750c1bf9

marcoroth (approved):
This looks good me now! Thanks @seanpdoyle!

Merry646 (commented):
Hello 👋How are things going on with you

PR #643, opened 2023-01-24
Explicitly type options of `Controller.dispatch()` method

### LINKED ISSUE #642: Allow dispatch target type to not be the same as the controller element type
### Summary

When using generics to declare a more specific element type on the controller, the `dispatch` target type is linked to that same type and it is no longer possible (without overrides or type checking warnings) to dispatch against a target of any kind of element.

Not sure if this is a bug but it is probably an unintended side effect of allowing a type signature for controllers via generics completed in https://github.com/hotwired/stimulus/pull/529

### Steps to reproduce

1. Create a new controller (TypeScript) similar to this, where the Controller's type is `HTMLInputElement` (e.g. not a generic `Element` or `HTMLElement`.
```
import { Controller } from '@hotwired/stimulus';

export class ExampleController extends Controller<HTMLInputElement> {
  connect() {
    const main = document.querySelector('main') as HTMLElement;

    this.dispatch('Connected!', { target: main });
  }
}
```
2. Expected: I should be able to use the documented `dispatch` callback with any target element - as per https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events (see `dispatch accepts additional options as the second parameter as follows:`)

3. Actual: I get a type error advising that the dispatch target element `HTMLElement` does not match the `HTMLInputElement` type.

```
Type 'HTMLElement' is missing the following properties from type 'HTMLInputElement': accept, align, alt, autocomplete, and 53 more.ts(2740)
```

A similar issue also happens if we were to pass in something like `document` to the target.

### Screenshots

Screenshot 2023-01-24 at 11 08 17 am
Screenshot 2023-01-24 at 11 10 16 am

### Potential solution 

Allow the `target` type to be something like:
```TypeScript
target?: ElementType | Element;
```

https://github.com/hotwired/stimulus/blob/eedb8ed2ff50fa5ad5319bc0e456d684046b2dbd/src/core/controller.ts#L82

```
 dispatch(
    eventName: string,
    { target = this.element, detail = {}, prefix = this.identifier, bubbles = true, cancelable = true } = {}
  ) {
    const type = prefix ? `${prefix}:${eventName}` : eventName
    const event = new CustomEvent(type, { detail, bubbles, cancelable })
    target.dispatchEvent(event)
    return event
  }
```

marcoroth:
This pull request explicitly types the options of the `Controller.dispatch()` method, so that the types don't need to be inferred.

In the case of #642 the type for the `target` option was just inferred from this `this.element`, which lead to the issue that you weren't able to dispatch the event on any other target element. It just worked if the target typed matched the generic type of the controller.

Resolves #642


src/core/controller.ts
```diff
@@ -8,6 +8,14 @@ import { ValuePropertiesBlessing, ValueDefinitionMap } from "./value_properties"
 
 export type ControllerConstructor = Constructor<Controller>
 
+type DispatchOptions = {
+  target?: Element | Window | Document
+  detail?: Object
+  prefix?: string
+  bubbles?: boolean
+  cancelable?: boolean
```
seanpdoyle, line 16: Instead of marking each property as optional, what do you think about changing `dispatch` to accept a `Partial<DispatchOptions>`?

marcoroth: Even better, thanks!

lb-:
Wow thanks @marcoroth  super fast. Also, I did not know about Partial - really appreciate it @seanpdoyle

PR #647, opened 2023-01-28
Add fallback selector for outlets

marcoroth:
This pull request adds a fallback selector for outlets so you don't have necessarily provide a selector in the `data-[identifier]-[outlet name]-outlet="[selector]"` attribute.

The selector will default to `[data-controller~=[outlet name]]`.

So an example like:
```html
<div data-controller="result"  class="result">...</div>
<div data-controller="result"  class="result">...</div>
<div data-controller="search" data-search-result-outlet=".result">...</div>
```

With a controller like:
```js
// app/javascript/controllers/search_controller.js

  static outlets = ["result"]

  connect() {
    this.resultOutlets.forEach(...)
  }
}
```

Can now be rewritten as:
```html
<div data-controller="result">...</div>
<div data-controller="result">...</div>
<div data-controller="search">...</div>
```

Of course you can still declare a selector to refine the selection of outlets.


src/core/outlet_set.ts
```diff
@@ -41,7 +41,11 @@ export class OutletSet {
 
   getSelectorForOutletName(outletName: string) {
     const attributeName = this.schema.outletAttributeForScope(this.identifier, outletName)
-    return this.controllerElement.getAttribute(attributeName)
+    const selector = this.controllerElement.getAttribute(attributeName)
+
+    if (selector) return selector
+
+    return this.scope.controllerSelector
```
seanpdoyle, line 48: This is minor, but what do you think about turning this early/middle `return` into either a full conditional, a ternary, or maybe `??`?

```ts
if (selector) return selector
else return this.scope.controllerSelector

if (selector) {
  return selector
} else {
  return this.scope.controllerSelector
}

return selector ?
  selector :
  this.scope.controllerSelector

return selector ?? this.scope.controllerSelector
```

marcoroth: I like the idea of using the nullish coalescing operator here

marcoroth: I went with ` ||`, thanks!

src/tests/modules/core/outlet_tests.ts
```diff
@@ -348,7 +349,32 @@ export default class OutletTests extends ControllerTestCase(OutletController) {
...
+      alpha1.classList.contains("disconnected"),
+      `expected "${alpha1.className}" to contain "disconnected"`
+    )
+    this.assert.ok(
+      alpha2.classList.contains("disconnected"),
+      `expected "${alpha2.className}" to contain "disconnected"`
+    )
+  }
+
+  async "skip test outlet disconnected callback when the controlled element's outlet attribute is removed"() {
```
marcoroth, line 377: @seanpdoyle this test case introduced in #624 is now failing due to the nature what this pull request changes.

Just removing the attribute on the controller element would mean that the "fallback selector" would come into play and therefore nothing gets disconnected in this case, because the fallback selector still matches the outlet elements.

Would you agree that this test case is not needed anymore? I adapted this test case to what you see on [L52](https://github.com/hotwired/stimulus/pull/647/files#diff-e52aed23e55a78cd208d17fbb4697756f2375b5aa2027dd00f967ed930fc2eb4R352).

seanpdoyle: What if there were an "atomic" operation that removed the attribute and the referenced element together before an `await this.nextFrame`?

marcoroth: Good idea, we could adapt that case to test for this case. Even though I'm not sure if that _could_ happen in a real app.

marcoroth: Unskipped in https://github.com/hotwired/stimulus/commit/617a749bef67d1589f93cc6dc25675f5d00b3d9b, thank you!

seanpdoyle:
I think I've encountered a scenario where having a default selector (the way it's proposed here) could have destructive side effects.

It's common for me to introduce an `element` controller to expose Element-level methods. For example, `element#click` would invoke `Element.click`, `element#focus` would invoke `HTMLElement.focus`, and even `element#requestSubmit` for `HTMLFormElement.requestSubmit`. So long as it's a one-to-one mapping between Stimulus Action Name and Method Name, I don't hesitate to add it to the `element` controller.

Imagine another controller that retains access to an `element` controller through an outlet. Let's assume it carries out some logic, then does some cleanup by invoking `element#remove` through the Outlet Controller interface.

If I were to typo the `[data-other-controller-element-outlet]` attribute name, the action would incur some nasty side effects. It would switch from executing on one (or small collection) of elements to executing on _any_ `[data-controller~=element]` element on the page, of which there are many.

The side effects could range from (best case scenario) a bunch of `console.error` noise because the affected elements don't define method that correspond to the Stimulus Action name to (worst case scenario) element#remove` stripping out far-flung elements, or submitting every `<form data-controller="element">` on the page (including the one's to destroy records!).

Without this change, a typo'd Outlet selector would silently fail.

marcoroth:
@seanpdoyle I feel like this is kind of an "edge case" scenario. Of course you need to be careful about spelling the attribute right, but I would also argue that this could be the case for other Stimulus APIs, depending what kind of actions you perform in your controller actions. But agreed, maybe it wouldn't be as "destructive" as in your example in those cases.

I feel like defining the attribute and leaving it without a value for it to fallback to the `[data-controller~=identifier]` selector also feels kinda weird. Like: 
```html
<div data-[identifier]-[outlet]-outlet></div>
```

Because that would still require you to "write out and declare" the attribute, which is what we want to optimize with this change.

Do you have an idea how we could improve this so that something simple like types wouldn't hurt that much?

flickgradley:
The ability to - by default - reference all other controllers of type _x_ on a page is super cool. It also matches how I tend to use outlets - I often invent a CSS class that I can use solely to find the other controllers, which have a name that matches the outlet.

I see the concern with the potential typo, though it does feel a bit edge-case-y to me too. Another idea is to make this opt-in - i.e. move towards options when declaring `outlets`:
```
static outlets = { result: { defaultSelector: true } }
```

PR #658, opened 2023-02-13
Ensure the `afterLoad` method gets called bound to controller

### LINKED ISSUE #652: Issue with using `this` inside `afterLoad` static method
### Problem

The afterLoad method is confusingly not called with a `this` context of the controller constructor but is instead undefined.

TypeScript is advising that `this` will be the controller constructor but at runtime it is failing.

### Steps to reproduce

Create a controller

```ts
export class LoaderController extends Controller {
  static values = {
    durationSeconds: { type: Number, default: 30 },
  };

  static afterLoad(identifier: string, application: Application) {
    const { controllerAttribute } = application.schema;
    console.log('afterLoad test', this); // this is undefined
    console.log('afterLoad stuff', {identifier, controllerAttribute });
    console.assert(this.values.durationSeconds.default, 30); // errors as this is undefined
  }
}
```

Expect that the `this` will be the controller constructor. Instead it is undefined.

### Potential solution

https://github.com/hotwired/stimulus/blob/8217e9b1224c5f8a82e9655574ed0d92173af74e/src/core/router.ts#L58

```ts
 const afterLoad = (definition.controllerConstructor as any).afterLoad?.bind(controllerConstructor)
```

lb-:
- Fixes #652

### DIFF
--- a/docs/reference/controllers.md
+++ b/docs/reference/controllers.md
@@ -169,13 +169,15 @@ If you want to trigger some behaviour once a controller has been registered you 
 
 ```js
 class SpinnerButton extends Controller {
+  static legacySelector = ".legacy-spinner-button"
+
   static afterLoad(identifier, application) {
     // use the application instance to read the configured 'data-controller' attribute
     const { controllerAttribute } = application.schema
 
     // update any legacy buttons with the controller's registered identifier
     const updateLegacySpinners = () => {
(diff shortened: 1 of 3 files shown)


marcoroth (approved):
Thank you @lb-!

lb-:
Thanks.

Issue #622, opened 2022-12-07
Value change callbacks are trigged by simply defining the value


fabian-as:
If I define a value without a default like
```javascript
static values = { "name": String }
```
the corresponding change callback `nameValueChanged()` will be triggered, even if the value is never set or changed.
I cobbled together a minimal working example
```html
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
    <script type="module">
      import { Application, Controller } from "https://unpkg.com/@hotwired/stimulus/dist/stimulus.js"
      window.Stimulus = Application.start()

      Stimulus.register("hello", class extends Controller {
        static values = { "name": String }

        nameValueChanged(oldName, newName) {
          console.log("Name was changed from", oldName, " to ", newName);
        }
      })
    </script>
  </head>
  <body>
    <div data-controller="hello"></div>
  </body>
</html>
```
which will print `Name was changed from <empty string>  to  undefined`.
Any chance on getting this fixed? Or is this the expected behavior?
I'm not quite sure, but I guess the reason is

https://github.com/hotwired/stimulus/blob/f690a6a2e856a35ebdb78246285b734da541338e/src/core/value_observer.ts#L76

where `defaultValue` should be checked against `null` too.

adrienpoly:
yes this is by design 

https://stimulus.hotwired.dev/reference/values#change-callbacks

> Stimulus invokes each change callback after the controller is initialized and again any time its associated data attribute changes. This includes changes as a result of assignment to the value’s setter.

marcoroth:
Yeah, this is by design so that you can react/initialize the initial state.

```js
static values = { "name": String }
```

is the shortcut for:
```js
static values = {
  "name": {
    type: String,
    default: ""
  }
}
```

And on connect it's going from `undefined` to `""` since that's the default value.

Similarly, if you define:
```html
<div data-controller="hello" data-hello-name-value="John"></div>
```

It would invoke the change callback once on connect with the new value being `"John"`

I hope this helps!

PR #661, opened 2023-02-17
Add registerDefaultEventName

### LINKED ISSUE #660: Feature request: allow extending `defaultEventNames` mapping
The logic for shorthand/default event names in stimulus is currently defined here: https://github.com/hotwired/stimulus/blob/f7bfc358e5df2cd4623ae47325254e4076d61148/src/core/action.ts#L89-L97

This is pretty handy, but if you are using a library of custom elements (e.g. shoelace) you have to turn `<button data-action="controller#method"></button>` into `<sl-button data-action="click->controller#method"></sl-button>`.

It's a tiny thing, but if stimulus were to simply export that mapping it would be possible for users to do something like:

```js
import { defaultEventNames } from '@hotwired/stimulus'

defaultEventNames['sl-button'] = () => 'click'
```

```html
<sl-button data-action="controller#method"></sl-button>
```

If exposing the mapping directly is considered to "open" a function like `setDefaultEventName(elementName, stringOrCallback)` would also be pretty nice.

If that makes sense to the maintainers I can take a crack at a PR.

grncdr:
This allows configure default events for arbitrary elements.

Closes #660

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -96,6 +96,14 @@ const defaultEventNames: { [tagName: string]: (element: Element) => string } = {
   textarea: () => "input",
 }
 
+export function registerDefaultEventName(tagName: string, eventName: string | ((element: Element) => string)) {
+  if (typeof eventName === 'string') {
+    defaultEventNames[tagName] = () => eventName
+  } else {
+    defaultEventNames[tagName] = eventName
+  }
+}
+
 export function getDefaultEventNameForElement(element: Element): string | undefined {
   const tagName = element.tagName.toLowerCase()
   if (tagName in defaultEventNames) {
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -1,3 +1,4 @@
+export { registerDefaultEventName } from "./action"
 export { ActionEvent } from "./action_event"
 export { Application } from "./application"
 export { Context } from "./context"
--- a/src/tests/modules/core/action_tests.ts
+++ b/src/tests/modules/core/action_tests.ts
@@ -1,4 +1,8 @@
+import { registerDefaultEventName } from "../../../core"
 import { LogControllerTestCase } from "../../cases/log_controller_test_case"
+
+// must be done before the fixture element is constructed
+registerDefaultEventName('some-element', 'click')
 
 export default class ActionTests extends LogControllerTestCase {
   identifier = "c"
@@ -14,10 +18,16 @@ export default class ActionTests extends LogControllerTestCase {
     <svg id="svgRoot" data-controller="c" data-action="click->c#log">
       <circle id="svgChild" data-action="mousedown->c#log" cx="5" cy="5" r="5">
     </svg>
+    <some-element data-controller="c" data-action="c#log">Hm</some-element>
   `
 
   async "test default event"() {
     await this.triggerEvent("button", "click")
+    this.assertActions({ name: "log", eventType: "click" })
+  }
+
+  async "test default event on custom element"() {
+    await this.triggerEvent("some-element", "click")
     this.assertActions({ name: "log", eventType: "click" })
   }
 


grncdr:
@marcoroth here's a first pass at implementing this feature. One thing that tripped me up was needing to ensure that the event name is registered _before_ stimulus attaches to the DOM and queries for elements. I guess in practice this would mean users need to call `registerDefaultEventName` before `Application.start()`.

Another thing I'm unsure about, is the "global" nature of this data; registering a default event name will effect all Application instances.

Given that Stimulus already has a `Schema` concept for extending keymappings and attribute naming conventions, I've also created a second PR (#662) that moves the `defaultEventNames` lookup into the schema. I don't have a strong preference for either approach.

adrienpoly (approved):
LGTM thanks for working on this 

It would be nice to have a bit of documentation. The example with shoelace component is great

marcoroth:
I think I prefer this solution here over the solution proposed in #662. In #622 you need to override the schema and need to make sure that you still handle the default events for the elements. 

With this solution proposed in this PR you can call `registerDefaultEventName()` for any element you like to add but don't need to think about the defaults already configured in Stimulus.

I guess the other solution would be to combine both approaches. So that you can call `registerDefaultEventName()` on the schema instance. That way it doesn't apply to all application instances since it's just scoped to each application.

grncdr:
> In https://github.com/hotwired/stimulus/issues/662 you need to override the schema and need to make sure that you still handle the default events for the elements.

The very first version of #662 was based on the way custom keyMappings are defined and looked like this:

https://github.com/hotwired/stimulus/blob/83e84b7e4cbbc60ae5cc55ae2ec26c04b03b9c32/src/tests/modules/core/action_custom_default_event_tests.ts#L7-L10

> I guess the other solution would be to combine both approaches. So that you can call `registerDefaultEventName()` on the schema **instance**.  That way it doesn't apply to all application instances since it's just scoped to each application/schema.

(emphasis mine)

The problem here is that (by default) all applications share a single `defaultSchema` instance.

This makes `schema.registerDefaultEventName()` _worse_ than a global `registerDefaultEventName` function IMO, because it _looks_ local to the application but would modify the shared `defaultSchema`. I suspect this is why there isn't a more ergonomic `schema.registerKeyMapping()` API, because it would be far too easy to accidentally leak changes between application instances. Telling users to write nested object spreads is ugly, but avoids mutating shared global state.

---

What would you think of changing the default value of the `schema` argument to `Application.constructor` to do a deep copy of the `defaultSchema`? That would allow adding mutating methods the schema associated with an application without risking leaks between applications. The mutating methods could even be defined on `Application` itself, so that the `Schema` remains a mostly internal implementation detail.

Concretely, that could look like:

```js
const app = new Application()

app.registerDefaultEventName('some-element', 'click')
app.registerKeyMapping('at', '@')

app.start()
```

and inside `Application`:

```ts
class Application {
  ...
  
  registerDefaultEventName(tagName: string, eventName: string | ((el: Element) => string)) {
    // mutating is fine here, because application uses it's own copy of the default schema
    if (typeof eventName === 'string') {
      this.schema.defaultEventNames[tagName] = () => eventName
    } else {
      this.schema.defaultEventNames[tagName] = eventName
    }
  }
  
  registerKeyMapping(alias: string, key: string) {
    // same here, much easier to document, and easier for users
    this.schema.keyMappings[alias] = key
  }
  
  ...
}
```

marcoroth:
> What would you think of changing the default value of the `schema` argument to `Application.constructor` to do a deep copy of the `defaultSchema`?

Oh right, I didn't think that far tbh, I kinda just assumed that's what we are already doing right now. And that every application instance already copies the `defaultSchema`.

We definitely shouldn't modify the global `defaultSchema` object. That's why I think it makes sense to copy the schema whenever a new application instance gets instantiated, as you pointed out 👍🏼 

I also like your proposal of having `registerDefaultEventName(...)` on `application` itself. This also lines up with what we have with `registerActionOption()` already.

For now I would just focus on `registerDefaultEventName` in this pull request, we can look into adding `registerKeyMapping` in a separate pull request.

PR #662, opened 2023-02-17
Move defaultEventNames into Schema

### LINKED ISSUE #660: Feature request: allow extending `defaultEventNames` mapping

grncdr:
This allows applications to extend the mapping with support for custom elements.

Closes #660

### DIFF
--- a/src/core/schema.ts
+++ b/src/core/schema.ts
@@ -5,6 +5,7 @@ export interface Schema {
   targetAttributeForScope(identifier: string): string
   outletAttributeForScope(identifier: string, outlet: string): string
   keyMappings: { [key: string]: string }
+  defaultEventNames: { [tagName: string]: (element: Element) => string }
 }
 
 export const defaultSchema: Schema = {
@@ -29,6 +30,22 @@ export const defaultSchema: Schema = {
     // [0-9]
     ...objectFromEntries("0123456789".split("").map((n) => [n, n])),
   },
+  defaultEventNames: {
+    a: () => "click",
+    button: () => "click",
+    form: () => "submit",
+    details: () => "toggle",
+    input: (e) => (e.getAttribute("type") == "submit" ? "click" : "input"),
+    select: () => "change",
+    textarea: () => "input",
+  }
+}
+
+export function getDefaultEventNameForElement(element: Element, schema = defaultSchema): string | undefined {
+  const tagName = element.tagName.toLowerCase()
+  if (tagName in schema.defaultEventNames) {
+    return schema.defaultEventNames[tagName](element)
+  }
 }
 
 function objectFromEntries(array: [string, any][]): object {
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -1,6 +1,6 @@
 import { ActionDescriptor, parseActionDescriptorString, stringifyEventTarget } from "./action_descriptor"
 import { Token } from "../mutation-observers"
-import { Schema } from "./schema"
+import { getDefaultEventNameForElement, Schema } from "./schema"
 import { camelize } from "./string_helpers"
 import { hasProperty } from "./utils"
 
@@ -23,7 +23,7 @@ export class Action {
     this.element = element
     this.index = index
     this.eventTarget = descriptor.eventTarget || element
-    this.eventName = descriptor.eventName || getDefaultEventNameForElement(element) || error("missing event name")
+    this.eventName = descriptor.eventName || getDefaultEventNameForElement(element, schema) || error("missing event name")
     this.eventOptions = descriptor.eventOptions || {}
     this.identifier = descriptor.identifier || error("missing identifier")
     this.methodName = descriptor.methodName || error("missing method name")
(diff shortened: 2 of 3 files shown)


src/core/schema.ts
```diff
@@ -29,6 +30,22 @@ export const defaultSchema: Schema = {
     // [0-9]
     ...objectFromEntries("0123456789".split("").map((n) => [n, n])),
   },
+  defaultEventNames: {
+    a: () => "click",
+    button: () => "click",
+    form: () => "submit",
+    details: () => "toggle",
+    input: (e) => (e.getAttribute("type") == "submit" ? "click" : "input"),
```
lb-, line 38:
```suggestion
    input: (element) => (element.getAttribute("type") == "submit" ? "click" : "input"),
```

Flyby comment - I know this is copied from the original code but maybe we can make this more readable by using a nicer variable name here.

src/core/schema.ts
```diff
@@ -29,6 +30,22 @@ export const defaultSchema: Schema = {
...
+    select: () => "change",
+    textarea: () => "input",
+  }
+}
+
+export function getDefaultEventNameForElement(element: Element, schema = defaultSchema): string | undefined {
```
lb-, line 44: Maybe this event should be an instance or static method on the Application instead?

While we are moving things around. Gives even more control if needed on custom Applications.

For example, if we wanted to make the default event different for a specific controlled element we could access the controller identifier in the function by `elemet.hasAttribute(this.schema.controllerIdentifier,...);`

src/core/schema.ts
```diff
@@ -29,6 +30,22 @@ export const defaultSchema: Schema = {
     // [0-9]
     ...objectFromEntries("0123456789".split("").map((n) => [n, n])),
   },
+  defaultEventNames: {
```
lb-, line 33:
```suggestion
  defaultEventMappings: {
```

Since we use the term mappings already, might be nice to align with that naming?

lb- (commented):
Just a few thoughts, I think this would be an awesome feature and helps for mixed adoption of Web Components.

lb-:
This probably needs docs also

grncdr:
> This probably needs docs also

Definitely up for adding docs. Can I interpret your review as tacit support for this implementation vs. #661?

src/core/schema.ts
grncdr: That would also require the `Action` constructor to receive an Application instance where as it currently receives only the schema.

---

However, I do think it would be nice to make this a method on schema, so I've made that change.

So now instead of this:

```ts
import {defaultSchema, Application} from '@hotwired/stimulus'

const schema = {
  ...defaultSchema,
  defaultEventNames: {
    ...defaultSchema.defaultEventNames,
    'sl-button': () => 'click',
    'sl-select': () => 'change',
    'sl-input': () => 'input',
  }
}

const application = new Application.start(document.documentElement, schema)
```

Consumers would do this:

```ts
import {defaultSchema, Application} from '@hotwired/stimulus'

const customElementDefaultEventNames = {
  'sl-button': 'click',
  'sl-select': 'change',
  'sl-input': 'input',
}

const schema = {
  ...defaultSchema,
  getDefaultEventNameForElement: (element) => (
    defaultSchema.getDefaultEventNameForElement(element) ||
    customElementDefaultEventNames[element.tagName.toLowerCase()]
  )
}

const application = new Application.start(document.documentElement, schema)
```

src/core/schema.ts
grncdr: Obsoleted by https://github.com/hotwired/stimulus/pull/662/commits/2ddac5f652f5f42cfc7eeb1f912be6aefa8ef8e7

marcoroth (commented):
Thanks for proposing two solutions for issue https://github.com/hotwired/stimulus/issues/660, @grncdr!

In case we are going to combine this and #661 I left some small remarks in this pull request. Otherwise this looks good to me. Thanks for working on this.

src/core/schema.ts
grncdr: I ended up dropping https://github.com/hotwired/stimulus/commit/2ddac5f652f5f42cfc7eeb1f912be6aefa8ef8e7 so this is relevant again. I'm open to renaming it.

grncdr:
This now takes into account the feedback on #661 as well, setting on an `Application#registerDefaultEventNames` method.

I opted for the plural form of the name and passing an object because it felt nicer to write this:

```js
app.registerDefaultEventNames({
  'sl-button': 'click',
  'sl-select': 'change',
  'sl-input': 'input',
  // etc.
})
```

vs this: 

```js
app.registerDefaultEventName('sl-button', 'click')
app.registerDefaultEventName('sl-select', 'change')
app.registerdDefaultEventName('sl-input', 'click')
// etc.
```

... but I'm open to changing it back to `registerDefaultEventName`.

---

## Other concerns

There's a limitation (called out in the docs) where this API won't work when calling `Application.start()` as a static method. You need to register default event names before the application starts or it won't properly bind listeners. If somebody really wants to use the static form, overriding the `Schema` passed to `start` is still possible:

```js
Application.start(document.body, {
  ...defaultSchema,
  defaultEventNames: {
    ...defaultSchema.defaultEventNames,
    'sl-button': 'click',
    // etc.
  }
})
```

src/core/schema.ts
grncdr: I tried this, but it wasn't really that nice (see https://github.com/hotwired/stimulus/pull/661#issuecomment-1592881090).

Issue #659, opened 2023-02-16
Documentation variables could be more diverse? 


benoittgt:
Hello

This subject has been initially open [on discourse](https://discuss.hotwired.dev/t/documentation-variables-could-be-more-diverse/4767).  Feel free to close it if you want to keep this as a discussion.

I enjoy reading the documentation of Stimulus but sometimes I find it hard to understand how variables are passed when they are too similar. For example on “[outlet](https://stimulus.hotwired.dev/reference/outlets)” doc where in multiple case “result” could be different?

![](https://global.discourse-cdn.com/standard10/uploads/stimulusjs/original/2X/f/f90558f3927a300a17f79198f10eda0a63e5c845.png)

Same in [cross controller event](https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events). Where we don’t know what we can change at first sight and what needs to be similar. Like at first reading I was thinking ; what “copy” represent?

![](https://global.discourse-cdn.com/standard10/uploads/stimulusjs/original/2X/0/01fcbe39ddf139db03d683a3caa7884fc3c90894.png)

Also I rarely have variables with one word. Handling a mix of dash, underscore (Rails), cameCase can leads to easy mistakes. Maybe the doc could be more diverse on this part too in examples.

Thanks

marcoroth:
Hey @benoittgt, thank you for taking the time to open this issue!

I really appreciate your feedback on the documentation and I agree with you that it could benefit from more diverse examples to better showcase the appropriate usage of each API. 

At the same time I think it's important that we don't use too different or too specific examples. So that they are still comparable and showcase how the API is used, without having to understand the context/domain of the example first, if that makes sense?

I'd be more than happy to merge any suggestions that you may have in mind to improve the documentation and provide a better understanding of the APIs. Thanks for bringing this up!

Issue #641, opened 2023-01-23
Outlet can't have unique name


amkisko:
It's not clear from reference documentation that outlet name has to be same as referenced controller identifier.

Also for some reason logic behind matching `element` does not work properly. I tried using outlets, it recognises the controller by identifier (same way if I use `Stimulus.router.module_by_identifier.get`), but can't match the element (I did not dig deeper).

tmbv93:
Furthermore, since we require the outlet to have a controller with the same name as the outlet, isn't having a CSS selector a bit redundant? It would be simpler to just define the outlets like this:

Parent controller HTML:
`<div data-controller="search">`

Parent controller JS:
`static outlets = ['result']`

Outlet HTML:
`<div data-controller="result">`

If we used this syntax, it would mean that all elements with `data-controller="result"` would automatically become outlets, but that seems like a small trade-off for simpler syntax. 

If outlets didn't need to have a controller with a matching identifier, it would be much easier to see the use case for CSS selectors. You could have several different controllers become outlets 

Parent controller HTML: 
`<div data-controller="search" data-search-result-outlet=".result">`

Parent controller JS:
`static outlets = ['result']`

Outlet:
`<div data-controller="result somethingelse" class="result">`

A bit tangential, but: I think it would be neat if both `result_controller.js` and `somethingelse_controller.js` in this example became outlets. It would allow for a listener pattern, where several types of elements can define custom behavior in response to actions done in the parent controller.

I'm a bit new to using Outlets, so sorry if I'm just misunderstanding the Outlets API completely!

marcoroth:
We probably could also offer a way to rename the outlet name, so that it doesn't need to be the controller identifier of the outlet controller. Previously we had the option as an array:
```js
static outlets = ["results"]
```
But we could probably also support a syntax with an object. The example above would be a shorthand for:
```js
static outlets = {
  "result": "result"
}
```
But with this syntax you could rename the outlet name using `[outlet controller identifier]: [outlet name in the host controller]`:
```js
static outlets = {
  "result": "anotherNameForResult"
}
```

So a call like:
```js
this.resultOutlets
```
would become:
```js
this.anotherNameForResultOutlets
```

What do you think?

---

@tmbv93 I actually have a pull request in progress to solve the first part you are describing, so that it falls back to `[data-controller~={outlet}]` if you didn't provide a more specific selector for the outlet.

amkisko:
[...]

Issue #628, opened 2022-12-19
Feature request: polymorphic outlets


sfnelson:
Outlets require that the target controller name corresponds to the outlet name, e.g. if I write `data-source-target-outlet` then my target element must have `data-controller="target"`.

This makes sense as a technical requirement, e.g. if there are multiple controllers registered on the target then the source must be able to discriminate between them to choose the controller to receive messages.

The technical constraint appears to us to be preventing object oriented polymorphism, i.e. we must statically know the name of the controller in order to send it messages, and it gets baked into the code for the source controller.

An example of where polymorphism would be useful is a generic "trigger" controller, ala `aria-controls` that triggers a change in state for a modal/menu controller.

Source:
```
class TriggerController extends Controller {
  static outlets = ["controlled"]
  toggle() {
    this.controlledOutlet.toggle();
  }
}
```

Targets:
```
// menu, or modal, or accordion, etc
class MenuController extends Controller {
  toggle() {
    ...
  }
}
```

We might use the source for a sliding menu, a modal+scrim, a menu, etc – anywhere that `aria-controls` might be appropriate.

We have previously solved this problem using events, but there's quite a bit of boilerplate to this approach and were hoping that outlets could simplify this approach.

seanpdoyle:
Thank you for opening this issue!

> We might use the source for a sliding menu, a modal+scrim, a menu, etc – anywhere that `aria-controls` might be appropriate.

You might find https://github.com/hotwired/stimulus/pull/627 to be interesting. I'd appreciate feedback as far as developer ergonomics goes. I'm also curious how it compares with the Outlet mental model, and how the two might coexist.

sfnelson:
@seanpdoyle that looks like exactly what we're trying to do! We'll give it a try, unfortunately unlikely to be before the new year.

ianterrell:
I found this issue when looking to do something similar. In my case, rather than a one-to-one polymorphic relationship I wanted a heterogenous collection of controller outlets. This use case felt similar enough that I thought I would add a note here.

I didn't see how references would exactly work for what I wanted, and so I worked up a solution with a bridging controller, which I think would also work in cases with a one-to-one polymorphic relationship.

[...]

Issue #615, opened 2022-11-30
Outlets array expects kebab-cased names instead of camel-cased ones (like targets do)


wynksaiddestroy:
If you have a **target** name which consists of multiple words, you'll define your targets as follows:
```
static targets = [ "errorMessage" ]
```
An **outlet** name with multiple words, has to be defined the following way:
```
static outlets = [ "error-message" ]
``` 
If this is the intended behaviour, it would be nice to add a small notice to the documentation. Thanks!

marcoroth:
Hey @wynksaiddestroy, thanks for opening this issue!

This is a good point, and I think I agree with you that it's kinda unexpected if you look at how targets are defined.

I'm wondering why I did it that way. I guess because controller identifiers are also being written in kebab-case, and since outlets reference controller identifiers I probably went with that.

I guess you could expect both ways, depending on how you look at it.

Not sure how we should proceed here. I'd be fine with changing it. But that would probably also mean it's a breaking change.

marcoroth:
Good point! I'd be happy to change it, but would also love to hear feedback from others.

connorwbey:
Agree it should be changed for consistency. Might as well be now while the feature is new.

dhh:
Would be happy to ship a quick 3.2.2 that addresses this. If there was a way to detect kebab case, still have it work with a warning of depreciation, and then document and recommend camel going forward, that would be great.

minimul:
I like kabab-case and wish targets worked as outlets do it that regard – but yeah consistency overrules.

drjayvee:
This may be unrelated, but the docs leave me wondering whether you need to use the outlet controller name verbatim, or whether you're allowed to alias it.

```
<div class="stuff" data-controller="awesome-stuff"></div>

<div
  data-controller="search"
  data-search-awesome-stuff-outlet=".stuff"
>

<!-- is the following allowed as well -->
<div
  data-controller="search"
  data-search-stuff-outlet=".stuff"
>
```

Would the 2nd then be allowed to use
```js
class extends Controller {
  static outlets = ['stuff']
  declare readonly stuffOutlet: AwesomeStuffController
}
```

rupasix:
> Originally, outlet getter had this problem.
> However, it used to be discernible at the time of definition and it was predictable that the two would be mixed.

Concerning is, if you use controllers in directory, your outlet controller names in camelcase are simplified, eg: `search--form` and `search-form` will be `this.searchFormOutlet`.

[...]

PR #663, opened 2023-02-17
Make inner mechanism of outlet API more obvious in documentation

benoittgt:
Hello

This is the first contributions linked to https://github.com/hotwired/stimulus/issues/659

I found it was difficult to understand how `outlet` API was working at first sight. Especially when you have a specific CSS selector that does not match the controller name. 

Also I found it hard to deal with controller name with dash, so I make it more obious in the doc.

I know that the documentation looks less "Stimulus" style, but I think it will help dev to understand the inner mechanism of _outlet_. 

I use [jsfiddle](https://jsfiddle.net/2vgznf90/2/) as a playground. I was not able to reproduce the errors from the doc.

Related:
  - https://github.com/hotwired/stimulus/issues/615
  - https://github.com/hotwired/stimulus/issues/641
  - https://github.com/hotwired/stimulus/issues/628

### DIFF
--- a/docs/reference/outlets.md
+++ b/docs/reference/outlets.md
@@ -62,7 +62,7 @@ export default class extends Controller {
   static outlets = [ "user-status" ]
 
   connect () {
-    this.userStatusOutlets.forEach(userStatus => ...)
+    this.userStatusOutlets.forEach(status => ...)
   }
 }
 ```
@@ -106,7 +106,7 @@ export default class extends Controller {
   static outlets = [ "user-status" ]
 
   selectAll(event) {
-    this.userStatusOutlets.forEach(userStatus => userStatus.markAsSelected(event))
+    this.userStatusOutlets.forEach(status => status.markAsSelected(event))
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


gato-omega:
I actually had the same kind of issues understanding the docs since there's many ways the naming could be interpreted and some ambiguity on which specific naming conventions to use in certain situations (mostly the namespaced controllers and if the outlet could have a custom name or not, etc.). A bit of diversity in the examples/docs when the names are not "simple" could to the trick :)

marcoroth (approved):
Thank you for taking this on @benoittgt! 

I also think that it really helps to have distinct examples for each of the APIs, as you mentioned. So this looks good to me!

sedubois:
NB: this change hasn't yet been published to the website:
https://stimulus.hotwired.dev/reference/outlets

benoittgt:
Hello @marcoroth do you think the website could be refreshed? :)

marcoroth:
I can't do that, sadly. Maybe @dhh can trigger it?

dhh:
Done. Probably should move to the same method we have with Turbo where the docs just live in the website repo. I get how it's nice to tie docs and code together, but this syncing process is pretty inconvenient.

PR #665, opened 2023-03-07
Support keyboard modifier for click events

shouichi:
This commit adds functionality to accept modifiers for click events, e.g., `ctrl+click->link#open`.

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -4,6 +4,8 @@ import { Schema } from "./schema"
 import { camelize } from "./string_helpers"
 import { hasProperty } from "./utils"
 
+const allModifiers = ["meta", "ctrl", "alt", "shift"]
+
 export class Action {
   readonly element: Element
   readonly index: number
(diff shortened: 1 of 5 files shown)


lb-:
Just a thought, this can be done I think pretty easily with custom action options out of the box today. I am wondering of this would be sufficient for your requirements.

I have validated and the below works really well, it even supports the click shorthand data-action notation nicely.

```js

application.registerActionOption("alt", ({ event, value }) => value ? event.altKey : !event.altKey);
application.registerActionOption("ctrl", ({ event, value }) => value ? event.ctrlKey : !event.ctrlKey);
application.registerActionOption("meta", ({ event, value }) => value ? event.metaKey : !event.metaKey);
```

```html
<button type="button" data-action="gallery#view!alt gallery#expand:alt">View (using shorthand click action)</button>
<button type="button" data-action="click->gallery#expand:ctrl">View (only activate if ctrl is clicked)</button>
<button type="button" data-action="click->gallery#expand:alt:meta">View (only activate if both alt & meta are pressed)</button>
```

Related links: 

* https://stimulus.hotwired.dev/reference/actions
* https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

shouichi:
Cool! Do you think adding those custom action options to Stimulus itself is sensible? They seem common enough.

shouichi:
Thank you for the tip!

lb-:
Also. Should probably cover `shiftKey` also.

dhh:
Even if this is possible already by custom action options, I think it's considerably nicer to have part of the core package 👍

PR #675, opened 2023-04-24
Adding info about receiving controller

ThomasLandauer:
Took me forever to figure this out ;-)
Info is taken from https://fullstackheroes.com/tutorials/stimulus/create-custom-events/

Question: When exactly is it necessary to add `@window`?


adrienpoly:
> Question: When exactly is it necessary to add `@window`?

you need to add `@window` if your receiving controller is not the same node or a parent node of the emitting controller

PR #677, opened 2023-04-27
Add `PageUp` and `PageDown` to keyMappings

seb-jean:
I added `PageUp` and `PageDown` to keyMappings

### DIFF
--- a/src/core/schema.ts
+++ b/src/core/schema.ts
@@ -24,6 +24,8 @@ export const defaultSchema: Schema = {
     right: "ArrowRight",
     home: "Home",
     end: "End",
+    page_up: "PageUp",
+    page_left: "PageLeft",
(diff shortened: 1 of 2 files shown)


src/core/schema.ts
```diff
@@ -24,6 +24,8 @@ export const defaultSchema: Schema = {
     right: "ArrowRight",
     home: "Home",
     end: "End",
+    page_up: "PageUp",
+    page_left: "PageLeft",
```
lb-, line 28: Should this be `page_down: PageDown`?

dhh: Yes it should. Could you PR? Thanks for catching!

seb-jean: #678

Issue #638, opened 2023-01-17
Request: New dist/controller.min.js that only includes Controller.


dkniffin:
Over on https://discuss.hotwired.dev/t/advice-request-stimulus-webpack-best-practices-bundle-splitting-for-multi-page-ssr-site/612/12 we're discussing how to split apart stimulus controllers into separate packs in webpack. I think we've landed on a good way to do that:

```
// app/javascript/packs/globals.js
import { Application } from "@hotwired/stimulus";
window.Stimulus = Application.start();
```

```
// app/javascript/packs/foo_controller.js
import FooController from "controllers/foo_controller"
application.register("foo", FooController)
```
```
// app/javascript/packs/bar_controller.js
import BarController from "controllers/bar_controller"
application.register("bar", BarController)
```

This works. However, it means that Stimulus is loaded into each pack, increasing the size of each pack unnecessarily. The only thing we need from Stimulus in there is `Controller`.

To fix this, I want to request creating a new file `dist/controller.min.js` so that we can do `import { Controller } from '@hotwired/stimulus/dist/controller.js'`

dkniffin:
I'm able to workaround this by adding `window.Controller = Controller` to `global.js`, but that pollutes the global namespace, and I'd rather not do that.

marcoroth:
Hey @dkniffin, I just checked how that might work. Technically this is possible, but I'm wondering if it wouldn't make more sense if you instead tree-shake the bundle on your end.

Technically the tool should be smart enough to pull out the Constant and extract the related "dependencies" with it. Or is there any reason why you wouldn't want to go that route?

dkniffin:
@marcoroth maybe my app is configured wrong, but it's not tree-shaking everything else out currently. Do you know if there's a config for that in webpacker?

marcoroth:
@dkniffin did you have a chance to look at https://webpack.js.org/guides/tree-shaking?

dkniffin:
@marcoroth I took a look at that. I'll need to dig in more to my config and see if something is misconfigured. I'm definitely doing `import { Controller } from "@hotwired/stimulus"` so tree-shaking should work, I think.

But if I understand the webpack options correctly, only `usedExports` would shake out the unnecessary parts, right? I'm guessing `sideEffects` won't work, because all of Stimulus is in one file/module in the minified version. I suppose I could try unminified, then minify it as part of my build process, but that means I'd be (unnecessarily) building Stimulus every time.

marcoroth:
[...]

PR #686, opened 2023-06-15
Make sure to also run `build:test` in the `build` step

marcoroth:
This pull request makes sure that the `build:test` command is also run as part of the `build` and `release` step.

This way developers (especially plugin developers) can import all Stimulus' classes and types even if they are not publicly exported. We are not guaranteeing that those exports are non-breaking, but at least this will allow developers to make use of those classes even if we are not explicitly exporting them.

Importing classes would look something like:
```js
import { TargetSet } from "@hotwired/stimulus/dist/core/target_set"
import { Context } from "@hotwired/stimulus/dist/core/context"
```

And types could be imported like:
```ts
import type { TargetSet } from "@hotwired/stimulus/dist/types/core/target_set"
import type { Context } from "@hotwired/stimulus/dist/types/core/context"
```

Resolves https://github.com/hotwired/stimulus/pull/623 and https://github.com/hotwired/stimulus/issues/638

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,10 +31,11 @@
   "scripts": {
     "clean": "rm -fr dist",
     "types": "tsc --noEmit false --declaration true --emitDeclarationOnly true --outDir dist/types",
+    "prebuild": "yarn build:test",
     "build": "yarn types && rollup -c",
     "build:test": "tsc -b tsconfig.test.json",
     "watch": "rollup -wc",
-    "prerelease": "yarn build && git --no-pager diff && echo && npm pack --dry-run",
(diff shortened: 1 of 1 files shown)


DEfusion:
@marcoroth has this changed since this PR? With 3.2.2 only the types are included in the files directive of the package.

marcoroth:
Yeah, sorry for the confusion @DEfusion, this PR and the additional files in the NPM package are going to ship with the next release. 

If you need them today, please use the latest dev-build on `main`, thank you!

PR #693, opened 2023-06-20
Add support for number values with underscores

### LINKED ISSUE #646: Feature request - support numeric separators in number value attributes
## Summary

The ability to use an underscore as a separator within number values (e.g. `200_000`) is now part of the ECMA script spec and is supported in most browsers.

It would be great to see this supported in number values when used within Stimulus value attributes. 

For example, when wanting to support milliseconds but also ensure that the value is readable you could do the following.
```html
<button type="button" data-controller="wait" data-wait-delay-value="7_000">Delay for 7 seconds</button>
```

## Proposed implementation

The simplest way to resolve this would be to blanket remove underscore characters when parsing the number (example below). However, this is not explicitly aligned with the specification, but a regex could be written to be 100% aligned. 

https://github.com/hotwired/stimulus/blob/86564075f3b782f5ef6b5fbf875679444ab48280/src/core/value_properties.ts#L226

```ts
  return Number(value.replace('_','') 
```

The documentation would also need to be updated to reflect this support and explain the parsing logic. https://stimulus.hotwired.dev/reference/values#types

## Additional notes

While this is just syntactic sugar, it is backwards compatible and will help make long numbers more readable at a glance. 

When numbers are being updated by the controller itself, I assume there would be no attempt to re-apply the original formatting. This would only benefit manually written numbers in the html on first load / template responses.

## Links

- https://caniuse.com/mdn-javascript_grammar_numeric_separators
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number - `Number` does not support underscores in the provided string (note - neither does parseInt)
- http://ruby-doc.org/core-2.3.0/doc/syntax/literals_rdoc.html#label-Numbers - Ruby developers will be familiar with this syntax 
- https://peps.python.org/pep-0515/ - Python developers will also be familiar with this

lb-:
- Aligns with JavaScript numeric separators - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators (note: technically numeric separators are more strict but wanted to keep the parsing flexible & simple)
- Example `data-wait-delay-value="7_000"`
- Closes #646

### DIFF
--- a/docs/reference/actions.md
+++ b/docs/reference/actions.md
@@ -131,7 +131,7 @@ The list of supported modifier keys is shown below.
 
 Sometimes a controller needs to listen for events dispatched on the global `window` or `document` objects.
 
-You can append `@window` or `@document` to the event name (along with any filter modifier) in an action descriptor to install the event listener on `window` or `document`, respectively, as in the following example:
+You can append `@window` or `@document` to the event name (along with any filter modifer) in an action descriptor to install the event listener on `window` or `document`, respectively, as in the following example:
 
 <meta data-controller="callout" data-callout-text-value="resize@window">
 
--- a/docs/reference/values.md
+++ b/docs/reference/values.md
@@ -53,13 +53,13 @@ export default class extends Controller {
 
 A value's type is one of `Array`, `Boolean`, `Number`, `Object`, or `String`. The type determines how the value is transcoded between JavaScript and HTML.
 
-Type | Encoded as… | Decoded as…
----- | ----------- | -----------
-Array | `JSON.stringify(array)` | `JSON.parse(value)`
-Boolean | `boolean.toString()` | `!(value == "0" \|\| value == "false")`
-Number | `number.toString()` | `Number(value)`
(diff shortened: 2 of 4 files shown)


lb-:
Thanks for merging dhh

docs/reference/actions.md
```diff
@@ -131,7 +131,7 @@ The list of supported modifier keys is shown below.
 
 Sometimes a controller needs to listen for events dispatched on the global `window` or `document` objects.
 
-You can append `@window` or `@document` to the event name (along with any filter modifier) in an action descriptor to install the event listener on `window` or `document`, respectively, as in the following example:
+You can append `@window` or `@document` to the event name (along with any filter modifer) in an action descriptor to install the event listener on `window` or `document`, respectively, as in the following example:
```
adammiribyan, line 134: Is this a typo?

PR #695, opened 2023-06-20
Add support for `@outside` global event filter

### LINKED ISSUE #656: Feature request - Action option or global handling of 'outside' events
### Summary

The action options and global listeners are incredibly powerful. However, it would be great to fill in a common use case gap of an event that triggers only when outside the element is clicked.

Example use cases include focus traps, tooltips that close when something else it focused or clicked on, modals that can close when the user clicks outside the modal, dynamic form behaviour where blur is not suitable.

### Proposed usage

```html
<div data-controller="tooltip" data-action="click@outside->tooltip#close focus@outside->tooltip#close" data-tooltip-attached-to-value="#some-id">
  Tooltip content
</div>
```

```html
<div data-controller="focus-trap" data-action="focus@outside->focus-trap#keep keyup.tab->focus-trap#forward keyup.tab+shift->focus-trap#back">
  An overlay around a modal. 
</div>
```

### Alternative solutions

1. Don't implement this - instead it could be built as a custom event option, maybe just document it
2. Implementation via an action filter (like keyboard events) or a built in option (like prevent)

### Links

- Inspired by Alpine JS' `outside` directive https://enzedonline.com/en/tech-blog/wagtail-extending-the-draftail-editor-part-2-block-styles/
- https://github.com/hotwired/stimulus/issues/208 - this was closed, as the solution is quite simple, however since then there have been more Additions to the built in action syntax 
- https://github.com/stimulus-use/stimulus-use/blob/main/docs/use-click-outside.md - a util to add this behaviour to any controller
- https://discuss.hotwired.dev/t/best-practices-for-handling-clicks-outside-element/1266/5

lb-:
- When using `@outside`, it will behave the same as @document but only trigger the action if the event was triggered from outside the controlled element
- Closes #656


lb-:
**Important:** This has been built to consider the Controlled element as what should be the boundary, not sure on the best way to go about this. We could change this to be the bounds of the element that has the data-action to be more flexible.

For example, these will both behave the same, only clicks that are 'fully outside' the controlled element will call `a#something`.

```html
<main>
  <div id="outer" data-controller="a" data-action="click@outside->a#something">
    <button type="button" id="inner-button">outer<button>
    <div id="inner" data-action="click@outside->a#something">
      <button type="button" id="inner-button">inner</button>
    </div>
  </div>
  <button type="button">fully outside</button>
</main>
```

This may not be intuitive, so happy to adjust the PR so that it works based on the contains being compared against the element with the data-action.

src/core/binding.ts
```diff
@@ -82,6 +82,10 @@ export class Binding {
   private willBeInvokedByEvent(event: Event): boolean {
     const eventTarget = event.target
 
+    if (this.action.shouldIgnoreGlobalEvent(event, this.element)) {
```
lb-, line 85:
```suggestion
    if (this.action.shouldIgnoreGlobalEvent(event, this.action.element)) {
```

I believe this is all that will be needed to change the behaviour (see parent PR comment) so that the contains element is based on where the action was set up.

Also, I am not sure if I should be using something like the code below in this file.

```
this.scope.containsElement(this.action.element)
```

I am not 100% sure on the implications of using/not using the `this.scope`.

adrienpoly:
> This may not be intuitive, so happy to adjust the PR so that it works based on the contains being compared against the element with the data-action.

That would feel more natural to me having the outside being compared to the element with the data action and not the controller by itself.

lb-:
Not sure why it's failing, I don't think it's related to this PR. 

```
Safari 14.1 (Mac OS 10.15.7) EventOptionsTests custom option FAILED
	Expected: 2
	Actual: 1
	
	step
	fulfilled
	promiseReactionJob@[native code]
	
	Expected: {
	  "eventType": "toggle",
	  "name": "log"
	}
	Actual: {
	  "eventType": undefined,
	  "name": undefined
	}
	
	
	step
	fulfilled
	promiseReactionJob@[native code]
	
```

lb-:
Rebased

lb-:
Rebased

tonysm:
Nice work. I'd love to see this being merged! I was about to attempt implementing this after adding the same `closeWhenClickedOutside` action to different controllers.

lb-:
Just rebased

n-studio:
This feature is absolutely needed. I use `@click.outside` all the time when using Alpine.js.
When I switched to Stimulus it was the first thing I tried to do.

sjean-a:
> This feature is absolutely needed. I use `@click.outside` all the time when using Alpine.js. When I switched to Stimulus it was the first thing I tried to do.

Exactly

fritzmg:
`@outside` also makes sense to me - even more so with the ability to define a 'contains' element that is different from the controller's root as discussed above 🙂  👍

lukeify:
This would be a great feature to have in Stimulus, I'm sure I'm not the only one who inevitably ends up copying and pasting roughly the same "contains" logic checks into each of the apps I work on at some point. Time to make it a built-in feature!

Issue #672, opened 2023-04-16
how to add outlets to nested folders


pqvel:
i have 2 controllers which are on the path **controllers/create_controller.js** and **controllers/app/todos_controller.js**. How can I add outlet todos for the create controller?
_my HTML:_
![image](https://user-images.githubusercontent.com/76737719/232341771-abb853e6-c738-4f56-9e89-864465b85f6d.png)

_create_controller.js:_
`import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  static targets = ["todoText"]
  static outlets = ["todos"]

  createTodo(e) {
    e.preventDefault()
    this.todosOutlet.addTodo(this.todoTextTarget.value)
    this.todoTextTarget.value = ""
  }
}
`
_app/todos_controller.js:_
`import { Controller } from "@hotwired/stimulus"
import { v4 as uuid } from 'uuid'

export default class extends Controller {
  static targets = ["todos"]
  static values = { todos: Array }

  initialize() {
    this.todos = []
  }

  addTodo(todo) {
    this.todos.push({
        todo,
        id: uuid(),
        active: false
    })
    this.renderTodos()
  }

  renderTodos() {
    let todos = ''
    this.todos.forEach(todo => {
        todos += `
            <li class="${todo.active ? 'active' : ''}">
                <p>${todo.todo}</p>
                <input 
                    data-action="todos#changeActiveTodo" 
                    data-todos-id-param="${todo.id}"
                    type="checkbox"
                    value="${todo.active}"
                >
            </li>
        `
    })
    this.todosTarget.innerHTML = todos
  }
}
`

adiakritos:
I am also wondering the same thing. 

```
<div class="leads--foo" data-controller="leads--foo"></div>
<div data-controller="leads--table-filter" data-leads--table-filter-leads--foo-outlet=".leads--foo"></div>

// TABLE FILTER CONTROLLER

  static outlets = [ 'leadsFoo' ]

  connect(){
    console.log('table filter controller connected');
    console.log(this.hasLeadsFooOutlet);
    this.submitFormWhenDatePickerChanges()
    this.timeout = null;
  }            


// FOO CONTROLLER

export default class extends Controller {
  connect() {
    console.log('leads foo controller connected');
  }
} 

// CONSOLE OUTPUT
leads foo controller connected
table filter controller connected
false
```

fatk:
The issue is with the documentation being unclear about namespaced controllers. See #669 for details.

[...]

PR #696, opened 2023-06-23
Add nested outlet property documentation

swh-tropic:
Closes https://github.com/hotwired/stimulus/issues/672 by adding documentation on how to properly access nested controller outlets


raysuelzer:
This seems incorrect, or at least it lacks the html that should go with it. What would I use for this controller

query-builder--query_builder  (main controller)

outlet is for controller: `query-builder--rule-group`

What you have here does not work. The docs say explicitly to OMIT the namespace, but the example ADDs the namespace.

PR #697, opened 2023-06-26
Fixed value change observers not firing

### LINKED ISSUE #680: Change obeserver not called if controller name contains uppercase characters
I have encountered this behaviour that I believe to be a bug, where if a controller name contains an uppercase letter, [value change callbacks](https://stimulus.hotwired.dev/reference/values#change-callbacks) are never fired.

Everything behaves as expected after converting the controller name to lowercase.

Please see [this codepen](https://codepen.io/RB_/pen/gOBKxpO) for a demonstration: if you click the button, the text on the page does not change, however, if you rename the controller from `Test` to `test` (there are 4 occurrences in the html and 1 occurrence in the javascript) the text on the page changes after clicking the button. Tested on Firefox and Chrome.

RBastianini:
Fixed value change observers not firing when the controller name
contains upper case characters.

Fixes #680.

### DIFF
--- a/src/core/data_map.ts
+++ b/src/core/data_map.ts
@@ -43,6 +43,6 @@ export class DataMap {
   }
 
   getAttributeNameForKey(key: string): string {
-    return `data-${this.identifier}-${dasherize(key)}`
+    return `data-${this.identifier.toLowerCase()}-${dasherize(key)}`
   }
 }
--- a/src/tests/modules/core/value_tests_case_insensitive.ts
+++ b/src/tests/modules/core/value_tests_case_insensitive.ts
@@ -1,0 +1,92 @@
+import ValueTests from "./value_tests"
+
+export default class ValueCaseInsensitiveTests extends ValueTests {
+  identifier = "Test"
+
+  fixtureHTML = `
+    <div data-controller="${this.identifier}"
+      data-${this.identifier}-shadowed-boolean-value="true"
+      data-${this.identifier}-numeric-value="123"
+      data-${this.identifier}-string-value="ok"
+      data-${this.identifier}-ids-value="[1,2,3]"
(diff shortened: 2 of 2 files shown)


RBastianini:
I haven't found documentation about this, but it looks like the mutation observer reports property names always in lowercase (and since HTML is supposed to be case insensitive, I think it's fair). For this reason, my fix is to also track the properties that stimulus cares about in lowercase. I'm not sure the spot I touched is the right one, but at least no other test breaks.

src/core/data_map.ts
```diff
@@ -43,6 +43,6 @@ export class DataMap {
   }
 
   getAttributeNameForKey(key: string): string {
-    return `data-${this.identifier}-${dasherize(key)}`
+    return `data-${this.identifier.toLowerCase()}-${dasherize(key)}`
```
marcoroth, line 46: I'm not sure if this is the right fix for the issue raised in #680. The attribute methods on `Element` seem to be case-insensitive, which leads me to believe that we need to week this somewhere else.

RBastianini: If I get what you mean, they are case insensitive, meaning that you can read or write them with whatever casing and it would still work, but the mutation observer always reports the changed properties as lowercase strings. The consequence is that when we check if we should care about the property change, we always opt to discard the change event, because we instead stored the property with whatever casing was found in the document. I've not found this documented anywhere, but I tried both Chromium and Firefox and both did the same.

If I didn't get what you mean, could you point me in the direction where you think the fix should be made instead?

marcoroth: Hey @RBastianini, I just realized that my comment wasn't very well thought out, sorry about that! Let me dig into this and come back with a more constructive answer. Thank you!

PR #705, opened 2023-07-18
Update Values.md documentation

tellodaniel:
Updates: https://stimulus.hotwired.dev/reference/values

This PR aims to improve clarity for newcomers regarding the usage of the value data attribute. It emphasizes that the value data attribute should be defined in the same location as the controller. This clarification can help prevent unnecessary debugging efforts, as it took me around 15 minutes to resolve this issue personally.

### DIFF
--- a/docs/reference/values.md
+++ b/docs/reference/values.md
@@ -14,7 +14,7 @@ You can read and write [HTML data attributes](https://developer.mozilla.org/en-U
 </div>
 ```
 
-Data attributes used for values must be declared on the controlled element, the same one that has the `data-controller `attribute.
+Data attributes used for values must be declared on the controlled element, the same one that has the `data-controller` attribute.
 
 <meta data-controller="callout" data-callout-text-value="static values = { url: String }">
 <meta data-controller="callout" data-callout-text-value="this.urlValue">
--- a/docs/reference/values.md
+++ b/docs/reference/values.md
@@ -14,7 +14,7 @@ You can read and write [HTML data attributes](https://developer.mozilla.org/en-U
 </div>
 ```
 
-As per the given HTML snippet, remember to place the values for data attributes within the same element alongside the controller data attribute.
+As per the given HTML snippet, remember to place the data attributes for values on the same element as the `data-controller` attribute.
 
 <meta data-controller="callout" data-callout-text-value="static values = { url: String }">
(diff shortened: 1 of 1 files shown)


docs/reference/values.md
```diff
@@ -10,11 +10,12 @@ You can read and write [HTML data attributes](https://developer.mozilla.org/en-U
 <meta data-controller="callout" data-callout-text-value="data-loader-url-value=&quot;/messages&quot;">
 
 ```html
-<div data-controller="loader"
-     data-loader-url-value="/messages">
+<div data-controller="loader" data-loader-url-value="/messages">
 </div>
 ```
 
+The definition of the value data attribute should be specified in the same location as the controller data attribute.
```
lb-, line 17:
```suggestion
Data attributes used for values must be declared on the controlled element, the same one that has the `data-controller `attribute.
```

Just an idea, I think `must` is better than `should` here, also the concept of 'same location' could be confusing.

tellodaniel: Great suggestions. Thanks!

docs/reference/values.md
```diff
@@ -10,11 +10,12 @@ You can read and write [HTML data attributes](https://developer.mozilla.org/en-U
 <meta data-controller="callout" data-callout-text-value="data-loader-url-value=&quot;/messages&quot;">
 
 ```html
-<div data-controller="loader"
-     data-loader-url-value="/messages">
+<div data-controller="loader" data-loader-url-value="/messages">
 </div>
 ```
 
+Data attributes used for values must be declared on the controlled element, the same one that has the `data-controller` attribute.
```
marcoroth, line 17: I think we are mostly referring the element which holds the `data-controller`  attribute as the "controller element":
```suggestion
Data attributes used for values must be declared on the controller element, the same one that has the `data-controller` attribute.
```

tellodaniel: I believe we can omit the term `API` since the whole page is about Values. I will think we have enough context.

marcoroth: It's weird to have multiple different terms for the same thing though, especially if there is already an established one.

tellodaniel: I undestand. I change the text to a better one to include ideas from you comment and the previous one.

PR #715, opened 2023-08-20
Add support for `mod` key to be dynamic based on Mac/Windows

### LINKED ISSUE #654: Feature Request - Cross-platform "modify" for keydown filter.
I believe there is a common pattern to modify some keydown events using a different key on **Mac** and **Windows** operating systems. The one I see most commonly is using `CMD+Enter` on Mac and `CTRL+Enter` on Windows in order to **submit a form** while typing in a textarea.

I know that the `data-action` attribute on an element can take multiple actions; so, I could, in theory, achieve this as follows:

```html
<form data-controller="demo">
    <textarea data-action="
        keydown.meta+enter->demo#submit
        keydown.ctrl+enter->demo#submit
    "></textarea>
</form>
```

But, considering that I'd want to use this pattern in many places, this seems verbose.

My next thought was to register a custom keydown filter - `mod` - as follows:

```javascript
var macOsPattern = /Mac|iPod|iPhone|iPad/;

// Map the "mod" filter pattern to the OS-specific key.
defaultSchema.keyMappings[ "mod" ] = macOsPattern.test( window.navigator.platform )
	? "meta"
	: "ctrl"
;

window.Stimulus = Application.start( document.documentElement, defaultSchema );
```

... with the hope being that I could use something like `keydown.mod+enter` where it would be evaluated as `meta+enter` on Mac and `ctrl+enter` on Windows. But, unfortunately, it looks like modifier keys are handled _separately_ from the custom key mappings.

So, my thought would be to have Stimulus add a new modifier call `mod` or `modify` that would be the cross-platform evaluation.

Anyway, just throwing that out there. I raised it [on the Hotwired community forum](https://discuss.hotwired.dev/t/possible-to-map-mod-to-meta-on-mac-and-ctrl-on-windows/4768), and I tried Googling for it, and I didn't see anyone discussing this anywhere.

lb-:
- Adds `mod` to the `defaultSchema` which will resolve to `Meta` on a MacOS like device and `Control` on Windows like
- Allows `mod` to be overridden at the schema keyMappings level if needed and used as a standalone key
- Uses `mod` to be used as a key filter modifier for either `metaKey` or `ctrlKey` based on the resolved value
- Add unit tests and documentation for the `mod` key, and an example to the slideshow page
- Closes #654


lb-:
It appears that the CI is running as non-Mac but when running Chrome headless locally it runs as mac potentially.

I will have to review this PR a bit later and see if I can find a better way to test this feature in unit tests.

lb-:
Have adjusted the tests to provide better coverage of the different platforms, added some more tests and ensure the CI passes.

I have also added some questions about this PR to check the approach taken.

lb-:
Here is a branch comparison with one of the different approaches described above, instead of `mod` being inside `schema.keyMappings` it uses a memoized function to resolve the mod key.

This approach will give a bit more control to any overrides of the schema and avoids confusion where `mod` cannot be used as a standalone key even though it was going to be in `keyMappings`.

It does not solve for existing schema `mod` key mappings in the wild though which would no longer work with this approach. However, other schema usage could easily update their standalone mod usage to something like `$mod`.

https://github.com/hotwired/stimulus/compare/main...lb-:feature/654-mod-keyboard-modifier-support-alt-approach?expand=1

lb-:
Hey @marcoroth - hope it's ok to ping, just wondering if this could be reviewed. Thank you.

lb-:
Just rebased

PR #744, opened 2023-12-19
Removing self-closing tag

ThomasLandauer:


rupasix:
AFAIK, both cases, before and after that change, are not valid HTML. SPAN is not void element. https://developer.mozilla.org/en-US/docs/Glossary/Void_element

ThomasLandauer:
https://validator.w3.org/nu/ says:

> Warning: Self-closing tag syntax in text/html documents is [widely discouraged](https://google.github.io/styleguide/htmlcssguide.html#Document_Type); it’s unnecessary and [interacts badly with other HTML features](https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10809) (e.g., unquoted attribute values). If you’re using a tool that injects self-closing tag syntax into all void elements, [without any option to prevent it from doing so](https://github.com/prettier/prettier/issues/5246), then consider switching to a different tool.

In short:
`<span />` is wrong, whereas `<span>` is at least a valid starting tag.

rupasix:
I tried ...

marcoroth:
Let's make it right then and let us use `<span></span>`

ThomasLandauer:
OK, there you go :-)

PR #751, opened 2024-02-12
Removing rimraf as a dependency - not used anymore

thiagopradi:
Hi Folks,

This PR removes the package `rimraf` as a developer dependency since the package is not used anymore. 

Thiago


brunoprietog (approved):
Thanks @thiagopradi!

PR #760, opened 2024-03-19
Fix Inconsistency with Callback Ordering (#759)

### LINKED ISSUE #759: Callback Ordering Inconsistency in Documentation
The ordering of callbacks is inconsistent in the documentation. It'd be nice if they were consistent. One place notes `[name]TargetConnected` callbacks get called _before_ `connect`, and the other says it gets called _after_.

"Lifecycle" wise and setup, it'd make sense that the target callbacks happen _after_ `connect()` and _before_ `disconnect()`. Note, the disconnect callbacks are consistent on both pages.

## Reference > Lifecycle Callbacks
* https://stimulus.hotwired.dev/reference/lifecycle-callbacks#targets

This page states the following
> The `[name]TargetConnected()` lifecycle callbacks will fire _before_ the controller’s `connect()` callback.

## Reference > Targets
* https://stimulus.hotwired.dev/reference/targets#connected-and-disconnected-callbacks

This page states the opposite
> Stimulus invokes each element callback any time its target elements are added or removed after `connect()` and before `disconnect()` lifecycle hooks.

zacheryph:
This resolves #759 and ensures lifecycle callbacks are documented in the order they happen.
This was validated by simply having the 4 lifecycle callbacks print to console, and having a button to remove the controller element from the document.

Screenshot 2024-03-19 at 4 16 04 PM

### DIFF
--- a/docs/reference/lifecycle_callbacks.md
+++ b/docs/reference/lifecycle_callbacks.md
@@ -71,7 +71,7 @@ A connected target will later become _disconnected_ when either of the preceding
 * the element's `data-{identifier}-target` attribute is removed or modified
 * the document installs a new `<body>` element, such as during a Turbo page change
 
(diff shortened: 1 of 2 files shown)


brunoprietog (approved):
Thanks @zacheryph

PR #771, opened 2024-05-25
Do not ScopeObserver#scopeReferenceCounts when outlet is called before controller is connected

### LINKED ISSUE #763: disconnect() not called if outlet is accessed in connect()
I noticed that accessing an outlet in `connect()` causes the reference count of that controller to be 2 instead of 1 after start has finished. That means all actions which under normal circumstances cause the controller to be disconnected (like removing the `data-controller` attribute) don't cause the disconnect anymore.

The following code snippet illustrates the problem

```html
<!doctype html>
<html>
<head>
  <meta charset="utf-8">
  <script type="module">

    Stimulus.register("child", class extends Controller {
      connect() {
        console.log("Child connected")
      }
      disconnect() {
        console.log("Child disconnected")
        alert("works as expected")
      }
    })

    Stimulus.register("parent", class extends Controller {
      static outlets = [ "child" ]
  
      connect() {
        console.log("Parent connected")
        console.log(this.childOutlet) // <- works without this line 💥
      }
      disconnect() {
        console.log("Parent disconnected")
      }
      
      listChildren() {
        console.log(this.childOutlet)
      }
    });

    document.getElementById("remove").addEventListener("click", function() {
      console.log("Removing data-controller attribute");
      document.getElementById("child").setAttribute("data-controller", "");
    });
  </script>
</head>
<body>
  <div id="parent" data-controller="parent" data-parent-child-outlet="#child">
    <button data-action="click->parent#listChildren">List children</button>
  </div>
  <div id="child" data-controller="child"></div>
  <button id="remove">Remove data-controller=child attribute</button>
</body>
</html>
```

Setting a breakpoint in `ScopeObserver.elementMatchedValue()` and reloading the page you can see that it is hit twice for `child`. Clicking the remove button does _not_ trigger `disconnect()` to be called. Everything works as expected with the marked line commented out.

ildarkayumov:
Fixes https://github.com/hotwired/stimulus/issues/763

The problem occurs when Outlet is accessed before controller has been connected by Stimulus and somebody tries to disconnect such controller later.

in such case `ScopeObserver#elementMatchedValue` is called 2 times so `this.scopeReferenceCounts` is eq 2.
and during `ScopeObserver#elementUnmatchedValue` this counter `this.scopeReferenceCounts` remains >= 1 so `scopeDisconnected(value)` is not called

### DIFF
--- a/src/core/module.ts
+++ b/src/core/module.ts
@@ -31,8 +31,11 @@ export class Module {
 
   connectContextForScope(scope: Scope) {
     const context = this.fetchContextForScope(scope)
-    this.connectedContexts.add(context)
-    context.connect()
+
+    if (!this.connectedContexts.has(context)) {
+      this.connectedContexts.add(context)
+      context.connect()
+    }
   }
 
   disconnectContextForScope(scope: Scope) {
--- a/src/core/router.ts
+++ b/src/core/router.ts
@@ -79,7 +79,7 @@ export class Router implements ScopeObserverDelegate {
     const scope = this.scopeObserver.parseValueForElementAndIdentifier(element, identifier)
 
     if (scope) {
-      this.scopeObserver.elementMatchedValue(scope.element, scope)
+      this.scopeConnected(scope)
     } else {
       console.error(`Couldn't find or create scope for identifier: "${identifier}" and element:`, element)
     }


src/core/module.ts
```diff
@@ -31,8 +31,11 @@ export class Module {
 
   connectContextForScope(scope: Scope) {
     const context = this.fetchContextForScope(scope)
-    this.connectedContexts.add(context)
-    context.connect()
+
+    if (!this.connectedContexts.has(context)) {
```
ildarkayumov, line 35: we need this to prevent double `controller#connect`

sebastianludwig:
I can confirm that this PR resolves the initially reported issue in #763 🎉

ildarkayumov:
@marcoroth can you check this please?

nemesit:
Would be nice if this could be merged

JaredSharplin:
This caused a production bug for us, I would love to see this merged and new stimulus release. For now we are using this as a patch.

Issue #717, opened 2023-08-28
Feature request: Allow custom Stimulus outlet name 


janko:
## Problem

I have a use case where users can manage "permission groups", which can have multiple "attachments". I modeled this by having a `permission-group-form` controller that handles general logic and `permission-group-form--attachment` controller that handles logic within a single attachment row.

```html
<div data-controller="permission-group-form">
  <div data-controller="permission-group-form--attachment">...</div>
  <div data-controller="permission-group-form--attachment">...</div>
  <div data-controller="permission-group-form--attachment">...</div>
  ...
</div>
```

I want to create an `attachments` outlet on the `permission-group-form` controller. However, since Stimulus requires outlet names to match the controller name, I have to call the outlet `permission-group-form--attachment`. This means I have to declare it like this:

```html
<div data-controller="permission-group-form" data-permission-group-form-permission-group-form--attachment-outlet=".attachment">
  <div class="attachment" data-controller="permission-group-form--attachment">...</div>
  <div class="attachment" data-controller="permission-group-form--attachment">...</div>
  <div class="attachment" data-controller="permission-group-form--attachment">...</div>
  ...
</div>
```

And in the controller I have a lot of repetition as well:

```js
// app/javascript/controllers/permission_group_form_controller.js
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  static outlets = [
    "permission-group-form--attachment" // I want just "attachment"
  ]

  connect() {
    this.permissionGroupFormAttachmentOutlets // I want just "attachmentOutlets"
  }
}
```

## Proposed solution

I wanted to propose the ability to choose a custom name for outlets, and map them to the Stimulus controller. The API could look something like this:

```js
// app/javascript/controllers/permission_group_form_controller.js
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  static outlets = {
    attachment: "permission-group-form--attachment"
  }

  connect() {
    this.attachmentOutlets //=> [...]
  }
}
```

And then the DOM declaration would be a lot simpler as well:

```html
<div data-controller="permission-group-form" data-permission-group-form-attachment-outlet=".attachment">
  ...
</div>
```

DavidPetrasek:
Duplicate issue here #641

imanpalsingh:
Hey @janko, will this help https://github.com/hotwired/stimulus/pull/786?

janko:
Yes, that seems exactly like what I wanted 👍🏻

PR #786, opened 2024-09-07
Add ability to rename outlets

imanpalsingh:
This PR adds ability to rename outlets with the following api

```js
// chat_controller.js

  static outlets = [ "user-status", { "features--read-recipients--time": "time" }]

  connect () {
    this.userStatusOutlet // Existing outlet
    this.timeOutlet // Renamed from this.featuresReadRecioientsTimeOutlet.
  }
}

```

Let me know if this would be useful as a feature.

Thank you!

### DIFF
--- a/src/core/controller.ts
+++ b/src/core/controller.ts
@@ -2,7 +2,7 @@ import { Application } from "./application"
 import { ClassPropertiesBlessing } from "./class_properties"
 import { Constructor } from "./constructor"
 import { Context } from "./context"
-import { OutletPropertiesBlessing } from "./outlet_properties"
+import { OutletPropertiesBlessing, OutletRenameObject } from "./outlet_properties"
 import { TargetPropertiesBlessing } from "./target_properties"
 import { ValuePropertiesBlessing, ValueDefinitionMap } from "./value_properties"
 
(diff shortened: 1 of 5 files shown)


DavidPetrasek:
That's confusing. Why access both the original and renamed property? Why not just specify a single name (if needed) like proposed here #717 ?

imanpalsingh:
> Why access both the original and renamed outlet? That's confusing. Here's I believe a better solution #717

Thank you for checking it but It does not create outlet with original name.
`this.userStatusOutlet` is a normal outlet which is not renamed. I suggest you to have a look at https://stimulus.hotwired.dev/reference/outlets

janko:
Would love to have this merged 🤞🏻  Without this functionality, outlet names are tied controller names, which is very limiting.

I currently have a `select` controller for `<select>` boxes, and a `geofence` controller that contains a select box for radius. The inability to change the outlet name forces me use generic `selectOutlet` name instead of the desired `radiusSelectOutlet`. Additionally, I can't have different outlets for different elements that activate the same controller.

MatheusRich:
@marcoroth @brunoprietog thoughts on this one? This would be super helpful. Thank you!

PR #845, opened 2025-06-02
Always clean up event listener on binding disconnect

### LINKED ISSUE #838: eventListeners are not removed when removing an element within a controller
This is related to the memory leak work fixed as part of https://github.com/hotwired/stimulus/pull/592.

That issue fixed cleaning up eventListeners when the controller is removed/replaced within the page. However, there are cases where the element with an action can be updated _within_ the controller (such as an input or button changing in response to other actions in a form).

I've altered the example from that issue to reproduce the new issue:

```
<html>

<head>
  <script type="module">
      doClick() { process(500) }
    })
  </script>
  <script>
    function process(count) {
      let i = 0
      let handler = setInterval(function () {
        if (++i > count) {
          clearInterval(handler)
        } else {
          element = document.getElementById("replace-in-controller")
          element.replaceWith(element.cloneNode(true))
        }
      }, 1)
    }
  </script>
</head>

<body>
    <div id="replace-in-controller">
      To reproduce:
      <ul>
        <li>Check heap snapshot</li>
      </ul>
    </div>
  </div>
</body>

</html>
```

Comparing the heap before and after clicking this button, you can see the detached elements relating to event listeners:

Image

In this case it appears that we execute https://github.com/hotwired/stimulus/blob/c953af5c2338a48873de10426325507cfd615988/src/core/binding_observer.ts#L70 to clean up the action, which does not specify `clearEventListeners` so they are not removed.

dfritsch:
Updates the binding observer and dispatch logic to always clean up the event listeners.

Without this change, removing an element _inside_ a controller can cause the bindings on that element to be removed but the event listeners are left dangling. Even if the controller is eventually also removed, these events will never be cleaned up, so these elements will remain in memory indefinitely.

Based on the tests, the one downside of this is that if you adjust an event that was set to once and move it to a different order on the element, that action now has a different index so will disconnect the old binding and add a new binding. Previously this would find the old event listener that wasn't properly cleaned up and re-use it, keeping the "once" logic. Actions which are added on top of the existing action would keep the same event, so this still partly works.

`clearEventListeners` would be passed as `true` in all cases with this update, so the parameter is removed.

Fixes #838

### DIFF
--- a/src/core/binding_observer.ts
+++ b/src/core/binding_observer.ts
@@ -7,7 +7,7 @@ import { Token, ValueListObserver, ValueListObserverDelegate } from "../mutation
 
 export interface BindingObserverDelegate extends ErrorHandler {
   bindingConnected(binding: Binding): void
-  bindingDisconnected(binding: Binding, clearEventListeners?: boolean): void
+  bindingDisconnected(binding: Binding): void
 }
 
 export class BindingObserver implements ValueListObserverDelegate<Action> {
@@ -72,7 +72,7 @@ export class BindingObserver implements ValueListObserverDelegate<Action> {
   }
 
   private disconnectAllActions() {
-    this.bindings.forEach((binding) => this.delegate.bindingDisconnected(binding, true))
+    this.bindings.forEach((binding) => this.delegate.bindingDisconnected(binding))
     this.bindingsByAction.clear()
   }
 
--- a/src/core/dispatcher.ts
+++ b/src/core/dispatcher.ts
@@ -41,9 +41,9 @@ export class Dispatcher implements BindingObserverDelegate {
     this.fetchEventListenerForBinding(binding).bindingConnected(binding)
   }
 
-  bindingDisconnected(binding: Binding, clearEventListeners = false) {
+  bindingDisconnected(binding: Binding) {
     this.fetchEventListenerForBinding(binding).bindingDisconnected(binding)
-    if (clearEventListeners) this.clearEventListenersForBinding(binding)
+    this.clearEventListenersForBinding(binding)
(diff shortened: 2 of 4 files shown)


src/core/binding_observer.ts
```diff
@@ -72,7 +72,7 @@ export class BindingObserver implements ValueListObserverDelegate<Action> {
   }
 
   private disconnectAllActions() {
-    this.bindings.forEach((binding) => this.delegate.bindingDisconnected(binding, true))
+    this.bindings.forEach((binding) => this.delegate.bindingDisconnected(binding))
```
dfritsch, line 75: The actual "fix" is adding `true` to line 70 above. However, this makes all cases true, so instead the option is removed entirely.

jorgemanrubia:
🤖 Thanks for the diagnosis in #838 and for this fix — the root cause analysis is spot on, and we just hit the same leak at production scale in Basecamp (Turbo morphs churning action elements under long-lived controllers, pinning hundreds of detached subtrees per hour).

We've opened #877 as an alternative that takes a slightly more conservative path: instead of clearing unconditionally (which, as you noted in the description, trades away the `{ once: true }` firing-history preservation across action rewrites on live elements — covered by the "preserves once history" edge case test), it passes `clearEventListeners: !action.element.isConnected`. Detached elements — the leak case, including the `replaceWith` repro from #838 — get cleaned up, while connected elements keep their cached listener so the existing once-history behavior is unchanged. The full suite passes unmodified, with a new regression test for the leak.

Closing this one in favor of #877, which also carries `Fixes #838`. Thanks again for getting to the bottom of this first.

PR #850, opened 2025-07-17
Add an example for custom key mappings

pdx91:


docs/reference/actions.md
```diff
@@ -106,12 +106,25 @@ import { Application, defaultSchema } from "@hotwired/stimulus"
 
 const customSchema = {
   ...defaultSchema,
-  keyMappings: { ...defaultSchema.keyMappings, at: "@" },
+  keyMappings: { 
+    ...defaultSchema.keyMappings, 
+    at: "@",
+    slash: "/",
+    question: "?"
+  },
```
MatheusRich, line 114: IMO this doesn't add much. The example below could use the existing `at` mapping.

docs/reference/actions.md
```diff
@@ -106,12 +106,25 @@ import { Application, defaultSchema } from "@hotwired/stimulus"
...
 
 const app = Application.start(document.documentElement, customSchema)
 ```
 
+Then use them in your action descriptors:
+
+```html
+<div data-controller="search" 
+     data-action="keydown.slash->search#focus keydown.question->search#showHelp">
+</div>
```
MatheusRich, line 125:
```suggestion
<div data-controller="mentions"
     data-action="keydown.at->mentions#displaySuggestions"></div>
```

PR #853, opened 2025-07-25
Fix Chrome autocomplete incorrectly triggerring actions

### LINKED ISSUE #852: Chrome autocomplete incorrectly triggers actions with keydown events (regardless of KeyboardEvent filters)
When using a Stimulus action like:

```html
<button data-action="keydown.meta+k@window->search#open">
```

In **Google Chrome**, clicking an input’s **autocomplete suggestion** unexpectedly triggers the controller action even though no key is pressed.

It seems that Chrome dispatches an `Event` with `type === "keydown"`, but it is not an instance of `KeyboardEvent`. This happens regardless of the key filter (and using `@window` or not).

### Reproduction

Use this CodePen:  https://codepen.io/matheusrich/pen/xbwOOqm

#### Steps:

1. Click inside the input field and wait for the autocomplete list to appear.
1. Pick a suggestion from the autocomplete list.
1. ✅ Stimulus incorrectly triggers the action

### Workaround

On my controller, I'm using this to prevent this behavior for now

```js
if (event.type == "keydown" && !(event instanceof KeyboardEvent)) return
```

I'll submit a fix to for this on Stimulus.

Chrome Version 138.0.7204.169 (Official Build) (x86_64)

MatheusRich:
When using a Stimulus action like

    <button data-action="keydown.meta+k@window->search#open">

in Google Chrome, clicking an input's autocomplete suggestion unexpectedly triggers the controller action even though no key is pressed.

It seems that Chrome dispatches an Event with type === "keydown", but it is not an instance of KeyboardEvent. This happens regardless of the key filter.

Closes #852

### DIFF
--- a/src/core/binding.ts
+++ b/src/core/binding.ts
@@ -86,8 +86,13 @@ export class Binding {
   private willBeInvokedByEvent(event: Event): boolean {
     const eventTarget = event.target
 
-    if (event instanceof KeyboardEvent && this.action.shouldIgnoreKeyboardEvent(event)) {
-      return false
+    if (event.type === "keydown") {
+      // when accepting an autocomplete suggestion, Chrome dispatches a keydown event,
+      // but it isn't a KeyboardEvent instance
+      const keyboardEvent: KeyboardEvent = event instanceof KeyboardEvent ? event : new KeyboardEvent("keydown")
+      if (this.action.shouldIgnoreKeyboardEvent(keyboardEvent)) {
+        return false
+      }
     }
 
     if (event instanceof MouseEvent && this.action.shouldIgnoreMouseEvent(event)) {
--- a/src/tests/modules/core/action_keyboard_filter_tests.ts
+++ b/src/tests/modules/core/action_keyboard_filter_tests.ts
@@ -197,4 +197,12 @@ export default class ActionKeyboardFilterTests extends LogControllerTestCase {
     await this.triggerEvent(button, "jquery.a")
     this.assertActions({ name: "log2", identifier: "a", eventType: "jquery.a", currentTarget: button })
   }
+
+  async "test ignore events dispatched by autocomplete"() {
+    const button = this.findElement("#button10")
+    await this.nextFrame
+    await this.triggerEvent(button, "keydown", {})
+
+    this.assertNoActions()
+  }
 }


seanpdoyle:
@jorgemanrubia are you able to enable this PR's CI workflow?

src/core/binding.ts
```diff
@@ -86,8 +86,13 @@ export class Binding {
   private willBeInvokedByEvent(event: Event): boolean {
     const eventTarget = event.target
 
-    if (event instanceof KeyboardEvent && this.action.shouldIgnoreKeyboardEvent(event)) {
-      return false
+    if (event.type === "keydown") {
+      // when accepting an autocomplete suggestion, Chrome dispatches a keydown event,
+      // but it isn't a KeyboardEvent instance
+      const keyboardEvent: KeyboardEvent = event instanceof KeyboardEvent ? event : new KeyboardEvent("keydown")
+      if (this.action.shouldIgnoreKeyboardEvent(keyboardEvent)) {
```
seanpdoyle, line 93: The underlying [Action.shouldIgnoreKeyboardEvent](https://github.com/hotwired/stimulus/blob/422eb81fa6496d7e24c3983c63e74f3530367cd3/src/core/action.ts#L42-L63) function ultimately [accesses properties on the `KeyboardEvent` instance](https://github.com/hotwired/stimulus/blob/422eb81fa6496d7e24c3983c63e74f3530367cd3/src/core/action.ts#L103), so instantiating a new `KeyboardEvent` instance without those properties might lead to unexpected behavior.

As an alternative, what do you think about using typecasting instead?

```suggestion
     // When accepting an autocomplete suggestion, Chrome dispatches a keydown event that is not a KeyboardEvent instance
    if (event.type === "keydown" && this.action.shouldIgnoreKeyboardEvent(event as KeyboardEvent)) {
```

That assumes that Chrome's synthetic event _does_ have the properties required to behave like a KeyboardEvent. That might not be the case, but my hunch is that it'd behave better as the syntetic event than as an event that Stimulus constructs itself.

MatheusRich: My intention was to use a newly created keyboard event where all of those keys are blank (so nothing matches). Typecasting probably works, but that synthetic event doesn't have any of the KeyboardEvent properties (besides `type`). The code would be a lot cleaner, though.

Issue #712, opened 2023-08-03
Handle error when controller not found


tmaier:
Lets assume I am referencing a controller at `data-controller="does-not-exist"`, which does not exist.
This can happen due to a typo - or because the controller has not been loaded for some reason (e.g. not in the load path, caching issue etc.)

When I see it correctly, this would never get raised. Not as an exception. Not even when the debug mode is enabled.

What is the rigtht approach to get this noticable? I would prefer if I can see it in my console when debug mode is enabled and when this will be handled by `Application#handleError`.

tmaier:
I believe this is the culprit: https://github.com/hotwired/stimulus/blob/7b810ec9dc9f35899027752bc6c24e336dc3da64/src/core/application.ts#L80-L83

It just swallows (returns `null`), if no controller has been found

Ted-Tash:
@tmaier 

Hello, if it is OK with you, I would love to work on this. My coworkers and I were speaking of this exact issue, and I thought it would be a fun way to contribute. I was excited to see someone had already opened an issue to track it.

tmaier:
Hi @Ted-Tash, this would be great!

PR #864, opened 2026-01-08
Log `Controller Not Found` when controller is not found.

Ted-Tash:
related to #712 

This PR passes a missing controller name to the logging debugger and then returns `null`. Previously a missing controller simply returned `null` and swallowed the error, which made debugging difficult. Here we still swallow the error (no change in user facing function), but with logging to help the developers.

### DIFF
--- a/src/core/application.ts
+++ b/src/core/application.ts
@@ -79,7 +79,12 @@ export class Application implements ErrorHandler {
 
   getControllerForElementAndIdentifier(element: Element, identifier: string): Controller | null {
     const context = this.router.getContextForElementAndIdentifier(element, identifier)
-    return context ? context.controller : null
+    if (context) {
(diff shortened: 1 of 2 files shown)


src/core/application.ts
```diff
@@ -79,7 +79,12 @@ export class Application implements ErrorHandler {
 
   getControllerForElementAndIdentifier(element: Element, identifier: string): Controller | null {
     const context = this.router.getContextForElementAndIdentifier(element, identifier)
-    return context ? context.controller : null
+    if (context) {
+      return context.controller
+    } else {
```
tabuna: I think we don't need an `else` operator here. We can use early return instead.

Ted-Tash:
@tabuna nudging for review 🙏

PR #869, opened 2026-02-20
Make key filters work on non-latin keyboard layouts

psd-coder:
## Overview
Fixes key filter matching in Stimulus actions so that keyboard shortcuts defined with ASCII characters (e.g., `keydown.j`) work correctly when the user's active keyboard layout produces non-ASCII characters (Cyrillic, Greek, Hebrew, etc.).

## Problem Statement
When users have a non-latin keyboard layout active (e.g., Russian phonetic, Greek, Hebrew), `event.key` contains the non-ASCII character produced by their layout rather than the Latin letter on the physical key. This caused key filters like `keydown.j->controller#action` to never match, making keyboard shortcuts unusable for non-latin keyboard users.

## Solution Approach
- Added `normalizedKeyboardEventKey()` that falls back to deriving the key from `event.code` when the mapped filter key is a printable ASCII character but `event.key` is not
- The fallback extracts the Latin letter or digit from `event.code` (e.g., `KeyJ` → `j`, `Digit3` → `3`) so physical key position is used instead of the produced character
- Fallback is skipped during IME composition (`event.isComposing`) to avoid interfering with input method editors
- Added tests for Cyrillic (Bulgarian phonetic), Greek, and Hebrew layouts

## Breaking Changes
None.

## Testing
- Three new test cases covering Cyrillic (й/KeyJ), Greek (ξ/KeyJ), and Hebrew (ח/KeyJ) keyboard layouts verifying the fallback works correctly

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -59,7 +59,10 @@ export class Action {
       error(`contains unknown key filter: ${this.keyFilter}`)
     }
 
-    return this.keyMappings[standardFilter].toLowerCase() !== event.key.toLowerCase()
+    const mappedKey = this.keyMappings[standardFilter].toLowerCase()
+    const normalizedEventKey = this.normalizedKeyboardEventKey(event, mappedKey)
+
+    return mappedKey !== normalizedEventKey
   }
 
   shouldIgnoreMouseEvent(event: MouseEvent): boolean {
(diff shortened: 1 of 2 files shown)


ai:
I want to bring more attention to this issue.

Right now, hotkeys are not working in 50% cases for all US-speaking users. For instance, in Fizzy it means that user can’t trust hotkey and stop using them.

PR #877, opened 2026-06-10
Clear cached event listeners when a removed action element leaves the document

### LINKED ISSUE #838: eventListeners are not removed when removing an element within a controller

jorgemanrubia:
## The leak

`Dispatcher#eventListenerMaps` is a strong `Map` keyed by EventTarget. Entries are only removed on the `disconnectAllActions` path — i.e. when a whole controller disconnects. When an element carrying `data-action` is removed while its controller element survives, `BindingObserver#disconnectAction` calls `bindingDisconnected(binding)` with `clearEventListeners` defaulting to `false`, so the `EventListener` (and the map entry keyed by the now-detached element) stays for the lifetime of the application — retaining the element and everything it references.

This pattern — descendants churning under a long-lived controller — is the normal outcome of morphing/partial re-renders (Turbo 8 morph refreshes, broadcast stream updates over lists). Profiling a production-sized Basecamp page receiving periodic morph re-renders of a list of ~70 rows showed hundreds of detached subtrees pinned exactly this way (each removed row's action button kept its whole row subtree alive, including nested `<turbo-frame>` elements and their observers), with retained heap growing ~135 KB per re-render, unbounded, in an otherwise idle tab. Heap snapshot retainer chain:

```
detached <button data-action="…">  ←(key)  Map (eventListenerMaps) ← Dispatcher ← Application
```

## The fix

`disconnectAction` now passes `clearEventListeners: true` when the action's element is no longer connected to the document. `Dispatcher#clearEventListenersForBinding` already guards with `hasBindings()`, so shared listeners survive while still in use.

The conditional matters: for elements that remain connected, the cached `EventListener` is intentionally preserved so `{ once: true }` firing history survives action attribute rewrites — the existing "edge case when updating action list with setAttribute preserves once history" test covers that and still passes. The cache is only dropped in the case where keeping it can no longer serve any purpose and instead leaks the element.

## Measurements (before/after, in the Basecamp reproduction)

Forced-GC measurements over 40 morph re-renders of the list, otherwise idle tab:

- Retained heap growth: **+5.4 MB → +1.0 MB**
- Retained detached action buttons: **272 → 0**
- Net `addEventListener` balance: **+264 → −280** (stale listeners now actually removed)

## Tests

Adds a regression test to `memory_tests.ts` (remove an action element while the controller remains → its cached listener is dropped). Full suite passes on Chrome and Firefox (430/430).

Fixes #838

### DIFF
--- a/src/core/binding_observer.ts
+++ b/src/core/binding_observer.ts
@@ -67,7 +67,12 @@ export class BindingObserver implements ValueListObserverDelegate<Action> {
     const binding = this.bindingsByAction.get(action)
     if (binding) {
       this.bindingsByAction.delete(action)
-      this.delegate.bindingDisconnected(binding)
+      // Clear the dispatcher's cached event listener when the action's element
+      // has left the document. Keeping the cache for connected elements
+      // preserves { once: true } firing history across action attribute
+      // updates, but caching listeners for removed elements would retain the
+      // element (and everything it references) for the application's lifetime.
+      this.delegate.bindingDisconnected(binding, !action.element.isConnected)
     }
   }
 
--- a/src/tests/modules/core/memory_tests.ts
+++ b/src/tests/modules/core/memory_tests.ts
@@ -19,4 +19,10 @@ export default class MemoryTests extends ControllerTestCase() {
     await this.fixtureElement.removeChild(this.controllerElement)
     this.assert.equal(this.application.dispatcher.eventListeners.length, 0)
   }
+
+  async "test removing an action element while its controller remains clears dangling eventListeners"() {
+    this.assert.equal(this.application.dispatcher.eventListeners.length, 2)
+    await this.remove("button[data-action$='#doLog']")
+    this.assert.equal(this.application.dispatcher.eventListeners.length, 1)
+  }
 }


marcoroth:
Is this related to https://github.com/hotwired/stimulus/issues/838 and #845?

jorgemanrubia:
🤖 Yes — thanks for pointing those out, @marcoroth! Both are the same underlying bug.

**#838 is exactly this leak.** An element carrying `data-action` removed *inside* a surviving controller goes through `disconnectAction` → `bindingDisconnected(binding)` with `clearEventListeners` defaulting to `false`, so the `eventListenerMaps` entry keyed by the now-detached element is never removed (#592 fixed the controller-removal path, but not this descendant-churn one). The `replaceWith(cloneNode)` repro in #838 is a synthetic version of what we measured in Basecamp at production scale — Turbo morph refreshes churning list rows under a long-lived controller, pinning hundreds of detached subtrees per hour. This PR fixes that repro, and I've added `Fixes #838` to the description.

**#845 had the same intent with an unconditional clear.** As its description notes, always clearing trades away the `{ once: true }` firing-history preservation across action attribute rewrites on live elements (the "edge case when updating action list with setAttribute preserves once history" test). This PR conditions the clear on `!action.element.isConnected` instead: detached elements — where the cached listener can never serve a purpose and only pins memory — get cleaned up, while connected elements keep their cache, so the once behavior is unchanged and the full suite passes unmodified. I've closed #845 in favor of this PR with a note crediting the original diagnosis.

